PR #38, opened 2017-12-12
Use EventEnriched in EventStreamBatch

mtesseract:
Use EventEnriched (including MetadataEnriched instead of plain Metadata) for event consumption.

When publishing events one uses the Metadata type, when consuming events one should use MetadataEnriched, right? This is after Nakadi has enriched the event's metadata.


mtesseract:
Feel free to give me +1 when this looks ok, @vitold.

vitold:
👍

PR #39, opened 2017-12-13
 Pluggable http backends

mtesseract:
This is for easier testing.

Especially the bug about not calling the request modifier on streaming can easily be turned into a test case when the Http Backend can be replaced with a dummy.

@etorreborre @vitold, maybe you would like to have a look at this.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -1,4 +1,4 @@
-version: '0.3.0.1'
+version: '0.3.1.0'
 synopsis: Client library for the Nakadi Event Broker
 description: This package implements a client library for interacting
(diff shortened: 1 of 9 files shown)


etorreborre:
Very nice, I will rebase my PR and do testing on top of that!

mtesseract:
Cool, I will merge this then to master.

etorreborre:
@mtesseract could you please try a similar test with this pluggable backend which reproduces what happens when streaming events?

Basically in my tests the first response I create seem to be used to create the conduit but this conduit never returns even if the body of the resource is finite. I don't understand how to properly simulate getting events for a subscription

PR #41, opened 2017-12-13
Make test suite runnable without integration tests

mtesseract:
Just some changes to the test suite so that it is more verbose and runnable without a local Nakadi server.


mtesseract:
Some more midnight commit @etorreborre, @vitold.

etorreborre:
That's nice, thanks.

vitold:
👍

PR #44, opened 2017-12-14
Exposed the constructor for the subscription stream context

amrhassan:
@etorreborre @mtesseract Let's discuss

### DIFF
--- a/src/Network/Nakadi/Types/Subscription.hs
+++ b/src/Network/Nakadi/Types/Subscription.hs
@@ -11,7 +11,7 @@ This module provides the Nakadi Subscription Types.
 -}
 
 module Network.Nakadi.Types.Subscription
-  ( SubscriptionEventStreamContext
+  ( SubscriptionEventStreamContext(..)
   ) where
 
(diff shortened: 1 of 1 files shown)


etorreborre:
@mtesseract I don't think we just need to expose the `StreamId` we already get it with `SubscriptionEventStream` when we start the subscription.
 
What I need is to create a `SubscriptionEventStreamContext` with the `StreamId`, `SubscriptionId` and the `Config`, all of which I already have, to be able to run the `ReaderT SubscriptionEventStreamContext`.

mtesseract:
Hi,

Ok. I understand that having this constructor exported is useful to you right now. Therefore, let's merge it.
At some point we should probably reconsider how to design the subscription streaming API in a way that is easier to adopt for non-conduit streaming. Maybe we can export a matching high-level API for other streaming packages side by side.

Background: My goal was to design the subscriptionSource/runSubscription/subscriptionSink API in such a way that the context can be completely opaque, it basically hides the work (streamId bookkeeping) necessary for commiting cursors.

Thanks!

PR #46, opened 2017-12-14
Use StrictData for all the data types

### LINKED ISSUE #42: Should all the datatypes be strict?
To prevent any unwanted memory leak. I don't have much experience on that front but it seems like a good preventive measure for such a library.

etorreborre:
fixes #42. I think this is the right thing to do


mtesseract:
Let's try it. :-) I don't remember anything in these ADTs where laziness was assumed.

PR #47, opened 2017-12-14
Cursor commit results

vitold:
Added request modifier into the streaming request.

### DIFF
--- a/src/Network/Nakadi/Internal/Http.hs
+++ b/src/Network/Nakadi/Internal/Http.hs
@@ -167,12 +167,14 @@ httpJsonBodyStream ::
   -> [(Status, ByteString.Lazy.ByteString -> m NakadiException)]
   -> (Request -> Request)
   -> m (b, ConduitM () a (ReaderT r m) ())
-httpJsonBodyStream config successStatus f exceptionMap _ = do
+httpJsonBodyStream config successStatus f exceptionMap requestDef = do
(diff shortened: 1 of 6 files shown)


src/Network/Nakadi/Internal/Http.hs
```diff
@@ -164,10 +170,11 @@ httpJsonBodyStream ::
 httpJsonBodyStream config successStatus f exceptionMap requestDef = do
   let manager        = config^.L.manager
       request        = requestDef (config^.L.requestTemplate)
-                       & setRequestManager manager
+                         & setRequestManager manager
       responseOpen   = config^.L.http.L.responseOpen
       responseClose  = config^.L.http.L.responseClose
-  (_, response) <- allocate (retryAction config (responseOpen request manager)) responseClose
+  modifiedRequest <- modifyRequest (config^.L.requestModifier) request
```
mtesseract: What is the reason for not using httpBuildRequest here?
This would also include setting the checkResponse field, which is important for exception handling.
This is responsible for letting 5xx responses throw exceptions, which in turn is important for the functioning of the retry handling.

vitold: Quite reasonable in terms of functional composition to use httpBuildRequest. I failed to do so on a first try, but now I see how to complete it.

mtesseract:
Cool, thanks!

PR #50, opened 2017-12-20
Use PartitionName type instead of Text

mtesseract:
@vitold, are you fine with this?
(also adding @etorreborre @amrhassan)


vitold:
@mtesseract sure

etorreborre:
👍

PR #52, opened 2017-12-20
Version bump

mtesseract:
Version bump, move tasty dependency to test section.


mtesseract:
nakadi-client builds just fine with lts-10.0, but it takes long on Travis CI, will take care of updating travis CI caches later to not delay the release any further.

PR #53, opened 2017-12-20
Implement devel flag cabal

mtesseract:
Hackage doesn't seem to accept packages which have -Werror set unconditionally.
This implements a flag 'devel' for the nakad-client package, which activates -Werror.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,4 +47,4 @@ after_script:
 - (cd $HOME/nakadi; TERM=dumb ./gradlew stopNakadi)
 
 script:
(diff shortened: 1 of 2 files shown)


mtesseract:
@etorreborre @vitold @amrhassan 
FYI

etorreborre:
I was totally unaware of that, maybe it is to avoid breaking package on innocent warnings?

mtesseract:
Yes, that seems to be the reason. New GHC versions may introduce new warnings, hence rendering such packages broken.

Issue #56, opened 2017-12-30
Generalizing callback monad and redesigning API


mtesseract:
## This is the first part of a major API redesign.

Note that the current API has never been declared stable.

I suggest to work on this for a future 0.5.0.0 release.
@etorreborre @vitold @amrhassan 

### What is wrong with the current API?

Several things are not perfect, but let's focus on one aspect for now. Namely, all callbacks registered by the user in the Nakadi configuration are to be run in the `IO` monad. This is not a strict limitation, but an inconvenience, as the user might be using some more complicated monad stack on top of `IO` for an application. Nakadi requiring callbacks in the `IO` monad therefore means that the user might has to do some type conversion, e.g. by unwrapping the custom monad stack, before providing the callbacks to nakadi-client.

### What would be nicer?

It would be nicer if the user could decide for some custom "base monad" in which the callbacks are to be run. As these callbacks are stored in the Nakadi configuration value `Config`, this value needs to be parameterized by this `base` monad. [Note, there is some conflict in terminology here. What is often called 'base monad' in Haskell (see e.g. the `transformers-base` package), is a specific monad among `STM`, `IO`, `Maybe`, etc..  Something like

    type App a = ReaderT IO a

would not be a 'base monad'. This is different in the usage of the word within nakadi-client. A base monad here means 'the custom monad transformer stack for your application'. Thus, `ReaderT IO` could be a base monad, if that is what the callbacks produce.

As mentioned previously, the `Config` type would now be parameterized by some base monad `b`. So, a nakadi-client function could look as follows:

```
eventTypesList :: SomeConstraints b => Config' b -> b [EventType]
```

But this would be a bad API, as this would mean that the only monad in which nakadi-client calls could be executed would be the application's base monad `b`. For example, consuming events, requires a `MonadResource` constraint on the monad to run in. More generally speaking, the user should always be free to locally run in some other monad stack — as long as it has the same base monad at the bottom. Otherwise we couldn't call the user's callbacks without unlifting them.

To sumarize the above: We need at least two monads `b` and `m` and a 'sub-monad' relationship between them. Thus, we end up with something like this: 
```
eventTypesList :: (MonadSub b m, ...) => Config' b -> m [EventType]
```

In the above, we actually need some more type class constraints on `b` and `m`, namely `MonadIO m`, `MonadIO b`, `MonadCatch m` and `MonadMask b`. 

Currently, the nakadi-client package exposes basically two sets of API calls:

[...]

PR #57, opened 2017-12-30
Parameterize monad

mtesseract:
#56 

I open the PR now to simplify discussion about it. We shouldn't rush this change.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -1,5 +1,5 @@
 name: nakadi-client
-version: '0.4.0.1'
+version: '0.5.0.0'
 synopsis: Client library for the Nakadi Event Broker
 description: This package implements a client library for interacting
              with the Nakadi event broker system developed by Zalando.
@@ -34,6 +34,8 @@ when:
     - -fno-warn-type-defaults
 dependencies:
 - base >=4.7 && <5
+- monad-control
+- transformers-base
 - conduit
 - conduit-extra
 - conduit-combinators
@@ -55,6 +57,7 @@ dependencies:
 - mtl
 - transformers
 - scientific
+- exceptions
 - safe-exceptions
 - unordered-containers
 - time
@@ -66,6 +69,7 @@ library:
   source-dirs: src
   exposed-modules:
   - Network.Nakadi
+  - Network.Nakadi.BackendIO
   - Network.Nakadi.Config
   - Network.Nakadi.EventTypes
   - Network.Nakadi.EventTypes.CursorDistances
--- a/src/Network/Nakadi/Config.hs
+++ b/src/Network/Nakadi/Config.hs
@@ -17,12 +17,11 @@ import           Network.Nakadi.Internal.Prelude
 
 import           Control.Lens
 import           Control.Retry
-import           Network.HTTP.Client             (Manager, ManagerSettings,
-                                                  responseClose, responseOpen)
-import           Network.HTTP.Client.TLS         (newTlsManagerWith,
-                                                  tlsManagerSettings)
-import           Network.HTTP.Simple             (httpLbs)
-import qualified Network.Nakadi.Internal.Lenses  as L
+import           Network.HTTP.Client               (Manager, ManagerSettings)
+import           Network.HTTP.Client.TLS           (newTlsManagerWith,
+                                                    tlsManagerSettings)
+import           Network.Nakadi.Internal.BackendIO
+import qualified Network.Nakadi.Internal.Lenses    as L
 import           Network.Nakadi.Internal.Types
 
 -- | Default retry policy.
@@ -32,12 +31,12 @@ defaultRetryPolicy = fullJitterBackoff 2 <> limitRetries 5
 -- | Producs a new configuration, with mandatory HTTP manager, default
 -- consumption parameters and HTTP request template.
 newConfig' ::
-  (MonadIO m, MonadThrow m)
+  MonadNakadi b m
   => Manager           -- ^ Manager Settings
   -> ConsumeParameters -- ^ Consumption Parameters
   -> Request           -- ^ Request Template
-  -> m Config          -- ^ Resulting Configuration
-newConfig' manager consumeParameters request =
+  -> m (Config' b)     -- ^ Resulting Configuration
+newConfig' manager consumeParameters request = do
   return Config { _consumeParameters              = consumeParameters
                 , _manager                        = manager
                 , _requestTemplate                = request
@@ -46,24 +45,17 @@ newConfig' manager consumeParameters request =
                 , _streamConnectCallback          = Nothing
                 , _logFunc                        = Nothing
                 , _retryPolicy                    = defaultRetryPolicy
-                , _http                           = defaultHttpBackend
+                , _http                           = backendIO manager
                 , _httpErrorCallback              = Nothing
                 }
-
--- | Default 'HttpBackend' doing IO using http-client.
-defaultHttpBackend :: HttpBackend
-defaultHttpBackend =
-  HttpBackend { _httpLbs                        = httpLbs
-              , _responseOpen                   = responseOpen
-              , _responseClose                  = responseClose }
 
 -- | Produce a new configuration, with optional HTTP manager settings
 -- and mandatory HTTP request template.
 newConfig ::
-  (MonadIO m, MonadThrow m)
+  MonadNakadi b m
   => Maybe ManagerSettings -- ^ Optional 'ManagerSettings'
   -> Request               -- ^ Request template for Nakadi requests
-  -> m Config              -- ^ Resulting Configuration
+  -> m (Config' b)         -- ^ Resulting Configuration
 newConfig mngrSettings request = do
   manager <- newTlsManagerWith (fromMaybe tlsManagerSettings mngrSettings)
   newConfig' manager defaultConsumeParameters request
@@ -85,7 +77,7 @@ setDeserializationFailureCallback cb = L.deserializationFailureCallback .~ Just 
 -- | Install a callback in the provided configuration which is used
 -- after having successfully established a streaming Nakadi
 -- connection.
-setStreamConnectCallback :: StreamConnectCallback  -> Config -> Config
+setStreamConnectCallback :: StreamConnectCallback m -> Config' m -> Config' m
 setStreamConnectCallback cb = L.streamConnectCallback .~ Just cb
 
 -- | Install a callback in the provided configuration which is called
@@ -93,7 +85,7 @@ setStreamConnectCallback cb = L.streamConnectCallback .~ Just cb
 -- conditions by e.g. logging errors or updating metrics. Note that
 -- this callback is called synchronously, thus blocking in this
 -- callback delays potential retry attempts.
-setHttpErrorCallback :: HttpErrorCallback -> Config -> Config
+setHttpErrorCallback :: HttpErrorCallback m -> Config' m -> Config' m
 setHttpErrorCallback cb = L.httpErrorCallback .~ Just cb
 
 -- | Install a logger callback in the provided configuration.
--- a/src/Network/Nakadi/Internal/Types.hs
+++ b/src/Network/Nakadi/Internal/Types.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE TypeFamilies           #-}
 {-|
 Module      : Network.Nakadi.Internal.Types
 Description : Nakadi Client Types (Internal)
@@ -10,9 +11,14 @@ Portability : POSIX
 Exports all types for internal usage.
 -}
 
-{-# LANGUAGE ConstraintKinds  #-}
-{-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE RankNTypes       #-}
+{-# LANGUAGE ConstraintKinds        #-}
+{-# LANGUAGE FlexibleContexts       #-}
+{-# LANGUAGE FlexibleInstances      #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE IncoherentInstances    #-}
+{-# LANGUAGE MultiParamTypeClasses  #-}
+{-# LANGUAGE RankNTypes             #-}
+{-# LANGUAGE UndecidableInstances   #-}
 
 module Network.Nakadi.Internal.Types
   ( module Network.Nakadi.Internal.Types.Config
@@ -23,11 +29,17 @@ module Network.Nakadi.Internal.Types
   , module Network.Nakadi.Internal.Types.Subscription
   , module Network.Nakadi.Internal.Types.Util
   , MonadNakadi
-  , MonadNakadiEnv
-  , HasNakadiConfig
+  , MonadNakadiEnv(..)
+  , MonadSub(..)
+  , NakadiT
+  , runNakadiT
   ) where
 
-import           Network.Nakadi.Internal.Lenses
+import           Control.Monad.Catch
+import           Control.Monad.Trans.Reader                 (ReaderT)
+import qualified Control.Monad.Trans.Reader                 as Reader
+import           Control.Monad.Trans.Resource
+import           Network.Nakadi.Internal.Prelude
 import           Network.Nakadi.Internal.Types.Config
 import           Network.Nakadi.Internal.Types.Exceptions
 import           Network.Nakadi.Internal.Types.Logger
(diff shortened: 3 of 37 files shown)


package.yaml
```diff
@@ -114,6 +118,9 @@ tests:
     - async
     - retry
     - stm
+    - stm-chans
+    - stm-conduit
     - warp
     - wai
     - tasty
+    - lifted-async
```
etorreborre: It would be better to have newlines at the end of a file: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

src/Network/Nakadi/Internal/Types/Config.hs
```diff
@@ -19,36 +19,39 @@ import           Network.Nakadi.Internal.Prelude
...
-type StreamConnectCallback = Maybe LogFunc -> Response () -> IO ()
+type StreamConnectCallback m = Maybe (LogFunc' m) -> Response () -> m ()
 
 -- | Type synonym for user-provided callbacks which are used for HTTP
 -- Errror propagation.
-type HttpErrorCallback = Request -> HttpException -> RetryStatus -> Bool -> IO ()
+type HttpErrorCallback m = Request -> HttpException -> RetryStatus -> Bool -> m ()
 
-data Config = Config
+type Config = Config' IO
```
etorreborre: A question about naming. I'm not such a big fan of `'` for exported names, I prefer that for local names. Why not use `Config` instead of `Config'` and `ConfigIO` for `Config' IO` (since we're breaking API anyway).

src/Network/Nakadi/Config.hs
```diff
@@ -46,67 +45,78 @@ newConfig' manager consumeParameters request =
...
-
 -- | Produce a new configuration, with optional HTTP manager settings
 -- and mandatory HTTP request template.
 newConfig ::
-  (MonadIO m, MonadThrow m)
+  MonadNakadi b m
   => Maybe ManagerSettings -- ^ Optional 'ManagerSettings'
   -> Request               -- ^ Request template for Nakadi requests
-  -> m Config              -- ^ Resulting Configuration
+  -> m (Config' b)         -- ^ Resulting Configuration
```
etorreborre: Also (see my below comment about naming): `newConfig` and `newConfig'` both return a `Config'` in this PR but the `'` are unrelated.

src/Network/Nakadi/Internal/Types.hs
```diff
@@ -36,17 +48,116 @@ import           Network.Nakadi.Internal.Types.Service
...
+  nakadiReader :: (Config' b -> a) -> m a
+  nakadiReader f = nakadiAsk >>= (return . f)
+
+instance (MonadIO m, MonadIO b, MonadCatch m, MonadMask b, MonadSub b m)
+      => MonadNakadiEnv b (ReaderT (Config' b) m) where
+  nakadiAsk = Reader.ask
+  nakadiLocal = Reader.local
+  nakadiReader = Reader.reader
+
+class (Monad b, Monad m) => MonadSub b m where
```
etorreborre: Is there nothing like that already in the Haskell eco-system? It feels like it should be something that Ed Kmett has already covered somewhere.

etorreborre: Maybe we can use this: https://hackage.haskell.org/package/natural-transformation-0.4/docs/Control-Natural.html?

mtesseract: I really would like to use natural-transformation, but I don't see an easy way to use it. It is basically built around the basic assumption, that both monads are fixed right. It then encodes a way to go from the one to the other.

But the use case in my PR is basically: There is a fixed base monad in the background (`b`) and that can be lifted to every monad transformer on top of that. So, we use a mechanism to go from `b a` to `m a` where `m` is not fixed. The caller might wants to use a ResourceT or a ReaderT or LoggingT or whatever on top of `b` and the nakadi-client API should just work.

I haven't found something like MonadSub anywhere. It's comparable to `MonadBase`, but with a more general notion of 'base'.

I guess I should write a blog post about this or something and ask the  community about it.

src/Network/Nakadi/Internal/Types/Config.hs
mtesseract: I agree completely.

src/Network/Nakadi/Config.hs
mtesseract: I agree completely. This was just done to reduce the breakage during development.

src/Network/Nakadi/Internal/Types.hs
etorreborre: Yes maybe it's worth asking on reddit, just to make sure we're covered.

mtesseract: Will do. Would you prefer the fooR -> foo change in the same PR? Actually it might make the PR a bit shorter, since it removes quite some lines.

etorreborre: Yes I think it's better.

package.yaml
mtesseract: Changed

src/Network/Nakadi/Internal/Types/Config.hs
mtesseract: Changed

src/Network/Nakadi/Config.hs
mtesseract: Changed

src/Network/Nakadi/Internal/Types.hs
mtesseract: done

mtesseract: Reddit: https://www.reddit.com/r/haskell/comments/7p2pc0/generalized_monadbase_monadsub_does_this_exist/
Let's see

src/Network/Nakadi/Config.hs
```diff
@@ -32,12 +31,12 @@ defaultRetryPolicy = fullJitterBackoff 2 <> limitRetries 5
...
-  (MonadIO m, MonadThrow m)
+  MonadNakadi b m
   => Manager           -- ^ Manager Settings
   -> ConsumeParameters -- ^ Consumption Parameters
   -> Request           -- ^ Request Template
-  -> m Config          -- ^ Resulting Configuration
-newConfig' manager consumeParameters request =
+  -> m (Config b)     -- ^ Resulting Configuration
+newConfig' manager consumeParameters request = do
   return Config { _consumeParameters              = consumeParameters
```
etorreborre: `do return` is just `return` or `pure` isn't it?

mtesseract: of course. :)

src/Network/Nakadi/Internal/Types.hs
```diff
@@ -36,17 +48,116 @@ import           Network.Nakadi.Internal.Types.Service
...
 
-import           Control.Exception.Safe
-import           Control.Monad.IO.Class
-import           Control.Monad.Reader.Class
-
--- | Type constraint synonym for encapsulating the monad constraints
+-- | type constraint synonym for encapsulating the monad constraints
 -- required by most funtions in this package.
-type MonadNakadi m = (MonadIO m, MonadCatch m, MonadThrow m, MonadMask m)
+type MonadNakadi b m = (MonadIO m, MonadIO b, MonadCatch m, MonadSub b m, MonadMask b)
```
etorreborre: Naming again, :-) ... 

 - `MonadNakadiEnv` is going to be used a lot more in the API than `MonadNakadi` and is longer to type

 - `MonadNakadi` is pretty abstract and not really about `Nakadi` while the `Config` part of `MonadNakadiEnv` is closer to the domain

So I think it would be nice to switch names and find another name for the current `MonadNakadi`, maybe `MonadSubControl` to draw an analogy with `MonadBaseControl`?

src/Network/Nakadi/Internal/Types/Logger.hs
```diff
@@ -17,4 +17,6 @@ import           Prelude
 
 -- | Type of a logger callback provided to nakadi-client for logging
 -- purposes.
-type LogFunc = Loc -> LogSource -> LogLevel -> LogStr -> IO ()
+type LogFunc' m = Loc -> LogSource -> LogLevel -> LogStr -> m ()
```
etorreborre: Also about naming, how about:
```
type LogFunc m = Loc -> LogSource -> LogLevel -> LogStr -> m ()
type LogFuncIO = LogFunc IO
```

src/Network/Nakadi/Internal/Types/Service.hs
```diff
@@ -803,3 +803,21 @@ data DataChangeEvent a = DataChangeEvent
   } deriving (Eq, Show, Generic)
 
 deriveJSON nakadiJsonOptions ''DataChangeEvent
+
+-- | Type of enriched DataChangeEvent.
+
+data DataChangeEventEnriched a = DataChangeEventEnriched
```
etorreborre: I thought we already had done a PR for this but apparently not. I think it would be better to keep this change outside of this PR. Am I missing anything?

src/Network/Nakadi/Internal/Types/Logger.hs
etorreborre: This also keeps the consistency with `ConfigIO`

tests/Network/Nakadi/Examples/Echo/Test.hs
```diff
@@ -0,0 +1,80 @@
...
+consumerMain eventName maxSize = runResourceT $ do
+  let consumeParameters = Nakadi.defaultConsumeParameters
+                          & L.batchFlushTimeout .~ Just 1
+                          & L.streamLimit .~ Just (fromIntegral maxSize)
+  source <- Nakadi.eventSource (Just consumeParameters) eventName Nothing
+  events <- runConduit $
+            source
+            .| concatMapC (view L.events)
+            .| concatC
+            .| mapC (id :: Nakadi.EventEnriched Foo -> Nakadi.EventEnriched Foo)
```
etorreborre: What's the purpose of this `mapC` with just `id`?

tests/Network/Nakadi/Examples/Echo/Test.hs
```diff
@@ -0,0 +1,80 @@
...
+            .| sinkVector
+  pure events
+
+testEcho :: Nakadi.Config IO -> Assertion
+testEcho config = Nakadi.runNakadiT config $ do
+  recreateEvent myEventTypeName myEventType
+  recreateEvent myEventTypeNameCopy myEventTypeCopy
+  events <- genEvents
+  let eventsPayloads = map (view L.payload) events
+  withAsync (runEcho myEventTypeName myEventTypeNameCopy) $ \ _echoHandle ->
```
etorreborre: `_echoHandle` is not interpreted as a type hole, why give it an underscore?

tests/Network/Nakadi/Tests/Common.hs
```diff
@@ -68,15 +75,18 @@ myDataChangeEvent eid now =  DataChangeEvent
...
-  eventTypeCreate conf eventType
-
-delayedPublish :: (ToJSON a) => Config -> Maybe FlowId -> [a] -> IO ()
-delayedPublish conf flowId events  = do
-  threadDelay (10^6)
-  eventPublish conf myEventTypeName flowId events
+genRandomUUID :: MonadIO m => m UUID
+genRandomUUID = liftIO randomIO
+
+recreateEvent :: MonadNakadiEnv b m => EventTypeName -> EventType -> m ()
```
etorreborre: I don't know where to leave this comment but I wonder if it would be worth moving some of the testing code we have for 'injecting' responses in the http backend here: https://github.bus.zalan.do/Glitch/sparrow/blob/acc69545260e4fb25cc3b7809f4512bc91236f3c/test/Test/Sparrow/Network/Nakadi/Config.hs (sorry people for the link to a private repository).

etorreborre:
I still haven't done a full thorough review but this is shaping good!

src/Network/Nakadi/Internal/Types.hs
mtesseract: Completely agree with MonadNakadiEnv → MonadNakadi.
I'm not even sure if we need the old MonadNakadi then. Currently it's used just a few times (... 2 times?!). I guess we could just drop the type synonym in this case.

etorreborre: Yes, we might as well remove it then.

tests/Network/Nakadi/Examples/Echo/Test.hs
mtesseract: Satisfying the type checker.
There are different types implement HasNakadiPayload as used in the next line.

src/Network/Nakadi/Internal/Types/Logger.hs
mtesseract: Done

tests/Network/Nakadi/Examples/Echo/Test.hs
mtesseract: This way there will no 'not used' warnings. Its like using "_" but plus some documentation advantage.

src/Network/Nakadi/Internal/Types.hs
mtesseract: done

src/Network/Nakadi/Config.hs
```diff
@@ -32,83 +30,85 @@ defaultRetryPolicy = fullJitterBackoff 2 <> limitRetries 5
...
-                , _httpErrorCallback              = Nothing
-                }
-
--- | Default 'HttpBackend' doing IO using http-client.
-defaultHttpBackend :: HttpBackend
-defaultHttpBackend =
-  HttpBackend { _httpLbs                        = httpLbs
-              , _responseOpen                   = responseOpen
-              , _responseClose                  = responseClose }
+  pure Config { _consumeParameters              = consumeParameters
```
etorreborre: Why use constraints on  `m` to eventually just do `pure`? `newConfig'` could just be a pure function creating a `Config b`. Then other functions can require `MonadIO m, MonadCatch m`.

src/Network/Nakadi/Internal/Types.hs
```diff
@@ -22,31 +32,285 @@ module Network.Nakadi.Internal.Types
...
+-- includes executing (non-streaming) HTTP requests and running
+-- user-provided callbacks. The typeclass provides methods for
+-- * retrieving the Nakadi configuration
+-- * locally changing the Nakadi configuration
+-- * extracting specific Nakadi configuration values
+-- * lifting actions from the
+-- The `MonadNakadi` typeclass is modelled closely after `MonadReader`.
+class (MonadNakadiHttp b m, MonadThrow m, MonadCatch m, NakadiHttpConstraint m)
+   => MonadNakadi b m | m -> b where
+  -- {-# MINIMAL (nakadiAsk | nakadiReader), local #-}
```
etorreborre: s/local/nakadiLocal ?

src/Network/Nakadi/Internal/Types.hs
```diff
@@ -22,31 +32,285 @@ module Network.Nakadi.Internal.Types
...
+  nakadiAsk :: m (Config b)
+  nakadiAsk = nakadiReader identity
+  nakadiLocal :: (Config b -> Config b) -> m a -> m a
+  nakadiReader :: (Config b -> a) -> m a
+  nakadiReader f = nakadiAsk >>= (return . f)
+  nakadiLift :: b a -> m a -- nakadiLift = lift
```
etorreborre: Is the comment intended to be at the end of the line?

src/Network/Nakadi/Internal/Types.hs
```diff
@@ -22,31 +32,285 @@ module Network.Nakadi.Internal.Types
...
+  type StM (NakadiT b m) a = ComposeSt (NakadiT b) m a
+  liftBaseWith = defaultLiftBaseWith
+  restoreM     = defaultRestoreM
+
+-- ** Implementations for `MonadNakadiHttp`.
+
+-- | `IO`-based implementation of `MonadNakadiHttp` for `NakadiT`
+-- using retries and `httpLbs`.
+instance (MonadIO b) => MonadNakadiHttp b (NakadiT b b) where
+  type NakadiHttpConstraint (NakadiT b b) = MonadMask b
```
etorreborre: I really had no clue that you could use type families like that to encode constraints on a given typeclass.

src/Network/Nakadi/Internal/Types.hs
```diff
@@ -22,31 +32,285 @@ module Network.Nakadi.Internal.Types
...
+  type NakadiHttpConstraint (ResourceT m) = NakadiHttpConstraint m
+  nakadiHttpLbs conf req mngr = lift $ nakadiHttpLbs conf req mngr
+
+-- | Inherit `MonadNakadiHttp` implementation to `ConduitM`.
+instance MonadNakadiHttp b m => MonadNakadiHttp b (ConduitM i o m) where
+  type NakadiHttpConstraint (ConduitM i o m) = NakadiHttpConstraint m
+  nakadiHttpLbs conf req mngr = lift $ nakadiHttpLbs conf req mngr
+
+-- | Inherit `MonadNakadiHttp` implementation to `LoggingT`.
+instance MonadNakadiHttp b m => MonadNakadiHttp b (LoggingT m) where
```
etorreborre: Worth adding `NoLoggingT` as well? Also note that we are proposing https://github.com/kazu-yamamoto/logger/pull/151 for tests.

src/Network/Nakadi/Subscriptions/Events.hs
```diff
@@ -30,39 +27,38 @@ import           Conduit
...
-  (MonadNakadi m, MonadResource m, MonadBaseControl IO m, MonadMask m, FromJSON a)
-  => Config                  -- ^ Configuration
-  -> Maybe ConsumeParameters -- ^ Optional Consumption Parameters
+  (MonadNakadiHttpStream b m, NakadiHttpStreamConstraint m, MonadResource m, MonadBaseControl IO m, FromJSON a)
+  => Maybe ConsumeParameters -- ^ Optional Consumption Parameters
   -> SubscriptionId          -- ^ Subscription ID
   -> m ( SubscriptionEventStream
        , ConduitM ()
          (SubscriptionEventStreamBatch a)
-         (ReaderT SubscriptionEventStreamContext m)
```
etorreborre: I welcome this simplification!

src/Network/Nakadi/Subscriptions/Events.hs
```diff
@@ -79,59 +75,10 @@ subscriptionSource config maybeParams subscriptionId = do
...
-  runSubscription config subscriptionEventStream conduit
-
 -- | Sink which can be used as sink for Conduits processing
 -- subscriptions events. This sink takes care of committing events. It
 -- can consume any values which contain Subscription Cursors.
 subscriptionSink ::
-  (MonadNakadi m, HasNakadiSubscriptionCursor a )
-  => ConduitM a Void (ReaderT SubscriptionEventStreamContext m) ()
+  (MonadNakadi b m, L.HasNakadiSubscriptionCursor a )
+  => ConduitM a Void (ReaderT (SubscriptionEventStreamContext b) m) ()
```
etorreborre: Is the sink going type signature going to align with `subscriptionSource`? Do you have some test code for this somewhere?

mtesseract:
Hi! I appreciate your comments a lot. Looks like I am overthrowing the complete type foundation at the moment, using some more MTL inspired techniques with any incoherent instances or the like.

So far it is looking good. I am currently working on another simplification: merging MonadNakadiHttp and MonadNakadiHttpStream.

How the actual streaming is exposed to the user might need some work as well, yes.

But right now I am rather optimistic, that we are on a good way.

I found this refactoring to be non-trivial, since the streaming is so heavily tight to IO.

mtesseract:
What I forgot so say: the PR is not completely done yet. :)

src/Network/Nakadi/Internal/Types.hs
mtesseract: Yeah, that is crucial for the current approach.

etorreborre:
I think what you are doing is more in line with Ed Kmett's tweet about how he would approach this. Also merging `MonadNakadiHttp` and `MonadNakadiHttpStream` seems right since both implementation are likely to be provided by the same library or module.

mtesseract:
See #59

Issue #68, opened 2018-01-23
Fails silently when streaming batches of un-enriched events


amrhassan:
Trying to stream objects that don't correspond to the structure of `EventEnriched a` fails silently and returns batches with `Nothing` in the `_events` field.

mtesseract:
For the record: the streaming APIs exposed by nakadi-client indeed (currently?) drop events that coudl not be deserialized. The reason behind this is that from a resilience perspective perspective a single event that failed to conform to some schema, should not bring down the whole service. Instead, what is additionally exposed is a mechanism to set a callback that is called on deserialization failures (`setDeserializationFailureCallback`). Using this, the user can handle the failure in whatever way that is desired.

It is debatable whether this is the right thing to do — maybe we should keep this issue open and exchange pros and cons?

Possible alternatives include:

  1. Fail and terminate stream consumption on deserialization failures
  2. Fail and terminate stream consumption EXCEPT if a deserialization failure callback is registered

Actually, I like the second idea. This would mean: Less surprises caused by silently ignoring errors. But IF the user knows about these callbacks, they can be installed and imply a different failure mechanism. Addtionally, using a return value, it could be decided if the failed events should simply be dropped or if some exception should be propagated upwards.

What do you think, @amrhassan ?

amrhassan:
I see your point about resilience, and I agree more about your second alternative of making it opt-in. I would always expect the default behaviour to not silently swallow up errors.

mtesseract:
This is done.

PR #70, opened 2018-01-25
Removed enrichment constraint from streamed type

amrhassan:
In effort to address the problem of not being able to stream events of a type other than `DataChangeEvent`s (as described in #68).


src/Network/Nakadi/Internal/Types/Service.hs
```diff
@@ -232,29 +232,6 @@ instance ToJSON EventId where
...
-  , _occurredAt :: Timestamp -- ^ Occurred-At timestamp
-  , _parentEids :: Maybe [EventId] -- ^ Event IDs of the Events which triggered this event
-  , _partition  :: Maybe PartitionName -- ^ Partition on which this Event is stored
-  } deriving (Eq, Show, Generic)
-
-deriveJSON nakadiJsonOptions ''Metadata
-
--- | Event
-
-data Event a = Event
```
amrhassan: I don't think this was used anywhere.

src/Network/Nakadi/Internal/Types/Service.hs
```diff
@@ -232,29 +232,6 @@ instance ToJSON EventId where
 instance FromJSON EventId where
   parseJSON = parseUUID "EventId" EventId
 
--- | Metadata
-
-data Metadata = Metadata
```
amrhassan: I've replaced this by the same name in the Nakadi doc, which is `EventMetadata`. I don't feel strongly about that though, I could revert it.

package.yaml
```diff
@@ -1,5 +1,5 @@
 name: nakadi-client
-version: '0.4.0.1'
```
mtesseract: Since this is an API breaking change, it should actually be 0.5. But since this is reserved already (kind of) for the upcoming version, I suggest to at least increment to 0.4.1.0.

src/Network/Nakadi/Internal/Types/Service.hs
```diff
@@ -721,40 +698,38 @@ data EventType = EventType
...
 
--- | Type of enriched event.
+-- | Type of received event metadata
 
-data EventEnriched a = EventEnriched
-  { _payload  :: a -- Cannot be named '_data', as this this would
-                   -- cause the lense 'data' to be created, which is a
-                   -- reserved keyword.
-  , _metadata :: MetadataEnriched
+data ReceivedEventMetadata = ReceivedEventMetadata
```
amrhassan: The received variant of `EventMetadata` which used to be `MetadataEnriched`. I could also revert the renaming here.

src/Network/Nakadi/Internal/Types/Service.hs
```diff
@@ -721,40 +698,38 @@ data EventType = EventType
...
-                                        , ("_metadata", "metadata") ]
-  }  ''EventEnriched
+deriveJSON nakadiJsonOptions ''ReceivedEventMetadata
 
 -- | EventStreamBatch
 
 data EventStreamBatch a = EventStreamBatch
   { _cursor :: Cursor -- ^ Cursor for this batch
-  , _events :: Maybe (Vector (EventEnriched a)) -- ^ Events in this batch
+  , _events :: Maybe (Vector a) -- ^ Events in this batch
```
amrhassan: The main change that allows streaming any kind of event type.

package.yaml
amrhassan: That's what I did :grin:

mtesseract: d'oh. Note to self: Don't read PRs when you need to catch a train.

src/Network/Nakadi/Internal/Types/Service.hs
```diff
@@ -721,40 +698,38 @@ data EventType = EventType
 
 deriveJSON nakadiJsonOptions ''EventType
 
--- | Type of enriched metadata values.
+-- | Type of published event metadata values.
 
-data MetadataEnriched = MetadataEnriched
+data EventMetadata = EventMetadata
   { _eid        :: EventId
   , _eventType  :: EventTypeName
```
mtesseract: Previously, `Metadata` did not contain an `EventTypeName` — should it according to the Nakadi API?
It seems redundant to include the event type name inside the event to be posted, since the event type name is already used as resource to which the POST request is sent.

Also, looks like this is the reason for the failing integration tests.

tests/Network/Nakadi/Tests/Common.hs
```diff
@@ -59,11 +59,12 @@ extractCursor Partition { ..} =
 myDataChangeEvent :: EventId -> UTCTime -> DataChangeEvent Foo
 myDataChangeEvent eid now =  DataChangeEvent
   { _payload = Foo "Hello!"
-  , _metadata = Metadata { _eid        = eid
-                         , _occurredAt = Timestamp now
-                         , _parentEids = Nothing
-                         , _partition  = Nothing
-                         }
```
mtesseract: I guess this line (66) is causing the test failures.

src/Network/Nakadi/Internal/Types/Service.hs
mtesseract: Ah, Nakadi says:

    If provided MUST match the endpoint. Failure to do so will cause rejection of the
    Event.
    
    (https://zalando.github.io/nakadi/manual.html#definition_Event)

I think this is something the library should handle for the user.

mtesseract:
Hi,

thanks again for the PR, in particular for fixing this bogus specialization to `DataChangeEvent a`.

But I am not sure about the renaming thing. After some initial concerns I thought it might be good from a self-documenting-code perspective. But, on the other hand, this naming might make code more confusing when the additional enriched fields are not actually used. Say, a user is not interested in the additional fields for some reason (an example might be an "Echo" service for republishing events).

If specially named *Received…* verions of the event data types exist, then I would expect them to be somehow "required" for receiving events. On the other hand *Enriched* is a little bit more transparent: These are just like the original versions but with some Nakadi-enriching added. But you can generally replace the *Enriched* versions with the *non-Enriched* versions if you don't need the added Nakadi metadata.

Does this make sense?

src/Network/Nakadi/Internal/Types/Service.hs
amrhassan: Indeed. Will drop again.

amrhassan:
Hey @mtesseract. Naming is hard :smiley:. I have no strong preference regarding the renaming, so I'm switching it back. I've kept the renaming of `Metadata` to `EventMetadata` to be the same as in the Nakadi doc though, but I can rename it back if you think against it.

src/Network/Nakadi/Types/Service.hs
```diff
@@ -65,9 +63,9 @@ module Network.Nakadi.Types.Service
   , PartitionKeyField(..)
   , EventType(..)
   , DataChangeEvent(..)
+  , DataChangeEventEnriched(..)
   , DataOp(..)
-  , EventEnriched(..)
-  , MetadataEnriched(..)
+  , EventMetadata(..)
```
mtesseract: I think we are missing an export for `EventMetadataEnriched(..)`, no?

mtesseract:
Hi,

I think this is looking good! I guess you are right that the `Metadata` →` EventMetadata` renaming makes the API closer to Nakadi's API. I'm not a big fan of doing such renamings in a minor version bump, but since you are probably the only user of this library, I guess this is fine. :-)

In the future we should probably use some deprecation method for this.

As I wrote, I think there is an export missing. But if you add this, I'm fine with merging.

Thanks!

src/Network/Nakadi/Types/Service.hs
amrhassan: Good eye. Will fix :)

mtesseract:
Thanks!

Issue #73, opened 2018-02-13
Implement MonadNakadi instance directly for ReaderT


mtesseract:
Actually `MonadNakadi` is nothing more than a specialized `ReaderT` monad. I had removed the `MonadNakadi` instance for `ReaderT` using some `HasNakadiConfig` constraint because I did not see how this would be when replacing replacing `Config` with `Config b`, because this `b` cannot be inferred anymore (... without tricks!).

I know understand that there is an obvious trick: functional dependencies. Using FDs it is straight forward to bring back the `HasNakadiConfig` typeclass and thus allow users to not use any other monad transformers besides straight `ReaderT`.

PR #74, opened 2018-02-13
Bring back HasNakadiConfig using Functional Dependencies and simplify MonadNakadi typeclass

mtesseract:
#73 

Define HasNakadiConfig typeclass using simplified interface and using functional dependencies.

Simplify MonadNakadi typeclass significantly. What nakadi-client itself only needs is a way to obtain the Nakadi configuration.
The other functionality that was inspired by ReaderT is not required within the library itself.

If a user needs this functionality (e.g., local), ReaderT together with the standard functions can be used.

Add test, which uses the MonadNakadi instance for ReaderT.

### DIFF
--- a/src/Network/Nakadi.hs
+++ b/src/Network/Nakadi.hs
@@ -24,7 +24,6 @@ import           Network.Nakadi.Base
 import           Network.Nakadi.Config
 import           Network.Nakadi.EventTypes
 import           Network.Nakadi.HttpBackendIO
-import           Network.Nakadi.Internal.Lenses
 import           Network.Nakadi.Registry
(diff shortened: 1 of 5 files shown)


mtesseract:
@etorreborre Here is a simplification that I had in my pipeline. :-)

I had only understood recently that using functional dependencies allows for saving the previous HasNakadiConfig mechanism.

etorreborre:
Indeed that's simpler, 👍. I need some to read this a bit better some day: https://web.cecs.pdx.edu/~mpj/pubs/fundeps-esop2000.pdf.

PR #77, opened 2018-02-15
Changed behaviour on deserialization failure

mtesseract:
If no callback is set, throw `DeserialiationFailure` exception.
Change this exception to also include a reason of failure, not only the bytestring which causes the failure.

This takes care of #68.

Can you review, @amrhassan ?

### DIFF
--- a/tests/Network/Nakadi/EventTypes/Test.hs
+++ b/tests/Network/Nakadi/EventTypes/Test.hs
@@ -123,7 +123,7 @@ testEventTypePublishData conf = runApp . runNakadiT conf $ do
                               }
   withAsync (delayedPublish Nothing [event]) $ \asyncHandle -> do
     liftIO $ link asyncHandle
-    eventConsumed :: Maybe (EventStreamBatch Foo) <-
+    eventConsumed :: Maybe (EventStreamBatch (DataChangeEvent Foo)) <-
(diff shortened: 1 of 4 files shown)


tests/Network/Nakadi/EventTypes/Test.hs
```diff
@@ -123,7 +123,7 @@ testEventTypePublishData conf = runApp . runNakadiT conf $ do
                               }
   withAsync (delayedPublish Nothing [event]) $ \asyncHandle -> do
     liftIO $ link asyncHandle
-    eventConsumed :: Maybe (EventStreamBatch Foo) <-
+    eventConsumed :: Maybe (EventStreamBatch (DataChangeEvent Foo)) <-
```
etorreborre: This change in the test doesn't test this new behaviour in the PR, right? In that case we should add a test showing that we indeed throw an exception.

etorreborre:
👍

tests/Network/Nakadi/EventTypes/Test.hs
amrhassan: I agree. A test would be great.

mtesseract: Right, this was fixing a test that was not working as expected.

I have fixed this and added a new test.

amrhassan:
:+1:

PR #79, opened 2018-02-15
Include FlowIds to all requests that accept X-Flow-Id header

mtesseract:
Add FlowId to all requests that expect a X-Flow-Id header as documented at
https://zalando.github.io/nakadi/manual.html
New function: setFlowId.

### DIFF
--- a/src/Network/Nakadi/EventTypes/EventType.hs
+++ b/src/Network/Nakadi/EventTypes/EventType.hs
@@ -50,8 +50,8 @@ eventTypeUpdate eventTypeName eventType = do
   config <- nakadiAsk
   httpJsonNoBody ok200 [] $
     (setRequestMethod "PUT"
-     . setRequestPath (path eventTypeName)
      . includeFlowId config
+     . setRequestPath (path eventTypeName)
      . setRequestBodyJSON eventType)
 
 -- | Deletes an event type given its 'EventTypeName'. @DELETE@ to
--- a/src/Network/Nakadi/EventTypes/EventType.hs
+++ b/src/Network/Nakadi/EventTypes/EventType.hs
@@ -28,14 +28,16 @@ path eventTypeName = "/event-types/" <> encodeUtf8 (unEventTypeName eventTypeNam
 
 -- | Retrieves an 'EventType' by its 'EventTypeName'. @GET@ to
 -- @\/event-types\/EVENT-TYPE@.
-eventTypeGet ::
-  MonadNakadi b m
+eventTypeGet
+  :: MonadNakadi b m
   => EventTypeName -- ^ Name of Event Type
   -> m EventType   -- ^ Event Type information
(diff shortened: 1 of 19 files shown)


mtesseract:
Would you like to review, @etorreborre ?
The FlowId handling was not correct.
I have included an X-Flow-Id header according to https://zalando.github.io/nakadi/manual.html.

src/Network/Nakadi/EventTypes/EventType.hs
```diff
@@ -44,18 +46,23 @@ eventTypeUpdate ::
...
-eventTypeUpdate eventTypeName eventType =
-  httpJsonNoBody ok200 []
-  (setRequestMethod "PUT"
-   . setRequestPath (path eventTypeName)
-   . setRequestBodyJSON eventType)
+eventTypeUpdate eventTypeName eventType = do
+  config <- nakadiAsk
+  httpJsonNoBody ok200 [] $
+    (setRequestMethod "PUT"
+     . setRequestPath (path eventTypeName)
```
etorreborre: This is a (very) minor thing but the code is not consistent w.r.t the order of calls `setRequestPath` and `includeFlowId`.

tests/Network/Nakadi/Tests/Common.hs
```diff
@@ -104,6 +104,9 @@ recreateEvent eventTypeName eventType = do
   eventTypeCreate eventType
 
 delayedPublish :: (MonadNakadi b m, MonadIO m, ToJSON a) => Maybe FlowId -> [a] -> m ()
-delayedPublish flowId events  = do
+delayedPublish maybeFlowId events  = do
   liftIO $ threadDelay (10^6)
-  eventsPublish myEventTypeName flowId events
+  let flowId = fromMaybe (FlowId "shalom") maybeFlowId
```
etorreborre: I like the flow id name :-)

etorreborre: Is there a way to test that we really emit the `X-Flow-Id` header? Or maybe add a unit test for `includeFlowId`?

etorreborre:
👍

src/Network/Nakadi/EventTypes/EventType.hs
mtesseract: True. I don't have strong feelings in this regard, but I have now moved the includeFlowId call to the top (before setRequestpath).

tests/Network/Nakadi/Tests/Common.hs
mtesseract: I have added tests.

etorreborre:
Time for the release or do you have anything else in mind?

mtesseract:
I hope I can do a release tonight.

mtesseract:
Thanks!

Issue #66, opened 2018-01-20
Smart / Fast Cursor Committing


mtesseract:
Currently, when consuming a subscription, every batch cursor is committed individually. This is not required and even more: For increasing throughput this is not even desirable. Better would be a smart cursor committing logic which skips some cursors, depending on event throughput and timings.

Such a "smart" subscription sink would probably have to be implemented as some seperate thread.

mtesseract:
Also, cursors on seperate partitions should be committed seperately. Furthermore, it looks like cursor commit failures should be discarded silently.

PR #86, opened 2018-02-23
Implement foundation for different cursor commit strategies

mtesseract:
New type: CommitStrategy
Include in Config
Default strategy reflects current behaviour: committing every cursor.
Can be extended with high-performance cursor committing strategies.
This is the first step towards #66.

@etorreborre @amrhassan

### DIFF
--- a/src/Network/Nakadi/Internal/Types/Subscriptions.hs
+++ b/src/Network/Nakadi/Internal/Types/Subscriptions.hs
@@ -13,8 +13,28 @@ subscriptions, which are not modelled after the Nakadi API.
 
 module Network.Nakadi.Internal.Types.Subscriptions where
 
+import           Data.Int
+
+
 -- | This type encodes the supported strategies for subscription
 -- cursor committing.
 data CommitStrategy
-  = CommitSyncUnbuffered -- ^ This strategy synchronously commit every
-                         -- cursor.
+  = CommitSync                          -- ^ This strategy synchronously commits every
+                                        -- cursor.
+  | CommitAsync CommitBufferingStrategy -- ^ This strategy sends cursors to be
+                                        -- committed to a dedicated thread
+                                        -- responsible for committing them. Cursors
+                                        -- are commited one by one, without special
+                                        -- buffering logic.
+
+-- | This type encodes the supported buffering strategies for
+-- asynchronous subscription cursor committing.
+data CommitBufferingStrategy
+  = CommitNoBuffer         -- ^ No buffering at all.
+  | CommitTimeBuffer Int32 -- ^ Buffer for the specified duration,
+                           -- given in milliseconds.
+  | CommitSmartBuffer      -- ^ Buffer for a fixed duration, but
+                           -- committing cursors immediately if the
+                           -- number of events processed since the
+                           -- last commit crosses a threshold derived
+                           -- from @maxUncommittedEvents@.
--- a/src/Network/Nakadi/Internal/Types/Subscriptions.hs
+++ b/src/Network/Nakadi/Internal/Types/Subscriptions.hs
@@ -1,0 +1,20 @@
+{-|
+Module      : Network.Nakadi.Types.Subscriptions
+Description : Nakadi Service Types
+Copyright   : (c) Moritz Clasmeier 2018
+License     : BSD3
+Maintainer  : mtesseract@silverratio.net
+Stability   : experimental
+Portability : POSIX
+
+This module defines types, which are related to consumption of
+subscriptions, which are not modelled after the Nakadi API.
+-}
+
+module Network.Nakadi.Internal.Types.Subscriptions where
+
+-- | This type encodes the supported strategies for subscription
+-- cursor committing.
+data CommitStrategy
+  = CommitSyncUnbuffered -- ^ This strategy synchronously commit every
+                         -- cursor.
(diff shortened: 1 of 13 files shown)


mtesseract:
This needs some changes prior to merging.

The naming is not correct (SyncUnbuffere  is in fact AsyncUnbuffered).

Will do this later.

src/Network/Nakadi/Internal/Types/Subscriptions.hs
```diff
@@ -13,8 +13,16 @@ subscriptions, which are not modelled after the Nakadi API.
...
+  = CommitNoBuffer
+
 -- | This type encodes the supported strategies for subscription
 -- cursor committing.
 data CommitStrategy
-  = CommitSyncUnbuffered -- ^ This strategy synchronously commit every
-                         -- cursor.
+  = CommitSync                          -- ^ This strategy synchronously commit every
+                                        -- cursor.
+  | CommitAsync CommitBufferingStrategy -- ^ This strategy sends cursors to be
```
etorreborre: I was wondering if there could be issues if you commit asynchronously and if committing fails? But I guess event consumption should really be idempotent so that doesn't matter.

etorreborre:
That looks good to me, maybe the hardest part is going to write a test for this feature :-). Also I was not aware that committing cursors could become a bottleneck. Did you have this experience in production?

src/Network/Nakadi/Internal/Types/Subscriptions.hs
mtesseract: In my proposal we are using withAsync to spawn a thread responsible for the cursor commiting. Actually, the crucial point here is that I have added a call to 'link' (line 133 in Events.hs). This causes any failure (read: exception) in the committer thread to be propagated to the parent thread.

This way, we can in fact propage failures.

mtesseract:
Regarding tests: It shouldn't be too hard, I guess. We have tests testing subscription consumption. We could run them for each cursor comitting strategy.

Regarding performance: Yes, if I remember correctly, it can have significant impact for high-throughput subscriptions. Doing an extra HTTP call for every batch and waiting for its response (potentially retrying) is relatively expensive. I believe this can be mitigated by tuning the batch size, though.

We should have benchmarks comparing the numbers.

src/Network/Nakadi/Subscriptions/Events.hs
```diff
@@ -25,32 +25,49 @@ module Network.Nakadi.Subscriptions.Events
...
   -> m ()
 subscriptionProcess maybeConsumeParameters subscriptionId processor =
   subscriptionProcessConduit maybeConsumeParameters subscriptionId conduit
   where conduit = iterMC processor
 
+-- | Conduit based interface for subscription consumption. Consumes
+-- the specified subscription using the commit strategy contained in
+-- the configuration. Each consumed event batch is then processed by
+-- the provided conduit. If an exception is thrown inside the conduit,
+-- subscription consumtion will terminate.
```
etorreborre: s/consumtion/consumption

src/Network/Nakadi/Internal/Types/Subscriptions.hs
etorreborre: Cool I didn't know about `link`.

etorreborre:
👍

mtesseract:
```
nakadi-client Test Suite
  Subscriptions
    Processing
      SubscriptionProcessing/async/TimeBuffer:                  OK (3.49s)
      SubscriptionProcessing/sync:                              OK (8.42s)
      SubscriptionProcessing/async/NoBuffer:                    OK (5.68s)
```

mtesseract:
Some updates on this.

Implementing this feature was pretty easy. But the test suite kept failing. :-(
It turned out to be tricky for me to debug, in particular since I
was not able to reproduce this problem on my machine.
In the end, it looks like the following was the problem:

One of my assumptions about the behaviour of the dockerized Nakadi
turned out to be wrong: I simply did not expect Nakadi to force
clients (i.e., the test suite) to reconnect by simply stopping
transmitting of non-empty batches. The tests for subscription consumption simply
stopped receiving batches with events at some point, receiving only
keep-alive cursors. Generally, it looks like the correct behaviour for
clients in such a situation is to let the connection terminate
gracefully. On the following reconnect Nakadi continues sending the
events on the stream and the test suite is happy.

This PR not only implements the foundation for smarter committing
strategies, but it also implements the smart committing strategy,
which is showing good performance in my tests.

I consider this now to be review-ready. Thanks. :-)
@etorreborre

etorreborre:
This looks all good to me even if I can not tell why your last commit is better than the previous one :-).

mtesseract:
Hi! There was a logical problem with the code before the last commit:

Using the smart strategy, committing is done on two occasions:

1. When some fixed duration has elapsed without commit
2. When at least one partition had crosses the configured event
   threshold.

On (2), the timer used for (1) was resetted and this is a per-stream
timer, not a per-partition timer(!). Thus, the following could happen:

Say we have two partitions, one being high-throughput (A), the other
one being very low-throughput (B) (artificial example, yes). Partition
(A) reaches the event threshold before the timer has elapsed. In that
case (before the commit!), only the cursor for partition (A) was
committed. Furthermore, if (A) had a constantly high event throughput,
always crossing the event threshold before the timer has elapsed, then
cursors for partition (B) would never be committed.

I see two solutions here:

 (i) Use per-partition committers including per-partition timer.

(ii) Simply commit all partitions when either the timer has elapsed or
     *some* partition has crossed the threshold.

While I think that (i) is, from a theoretical perspective, more
elegant, I think (ii) is sufficient for nakadi-client and much simpler
to implement (we would have to create and destroy timers dynamically,
depending on how many partitions Nakadi has asigned to use). That's
why I have quickly implemented (ii).

Issue #89, opened 2018-03-05
Fix event threshold configuration for smart cursor committing strategy


mtesseract:
For completeness, there is one problem with the current implementation: The smart committer (everybody should use this, should become the default at some point) derives its threshold from "maxUncommittedEvents".

What to do if that is Nothing? Currently it just picks some default, which is wrong, because this is not compatible with the actual configuration of the stream consumption.

So, everyhing should be sound if you use the smart committer PLUS you have explicitely set maxUncommittedEvents to some value, preferably large.

PR #90, opened 2018-03-05
Include event type in key in CursorsMap HashMap and fix Cursor Buffer Size Computation

mtesseract:
This should fix the case of having multiple event types behind a subscription.


mtesseract:
This should also take care of #89 
@etorreborre :-)

PR #93, opened 2018-03-07
Unliftio migration and remove low-level streaming API

mtesseract:
Preperation for upcoming changes regarding concurrency.
This PR makes nakadi-client adopt the new unliftio foundation and removes the deprecated (by Nakadi) low-level streaming API. See https://zalando.github.io/nakadi/manual.html#/event-types/name/events_get.


mtesseract:
@etorreborre, could you have a look when you find the time? :-)

tests/Network/Nakadi/EventTypes/Test.hs
```diff
@@ -122,18 +142,30 @@ testEventTypePublishData conf = runApp . runNakadiT conf $ do
...
                               }
-  withAsync (delayedPublish Nothing [event]) $ \asyncHandle -> do
+  subscriptionId <- createMySubscription
+  batchTv <- newTVarIO Nothing
+  res <- try $ withAsync (delayedPublish Nothing [event]) $ \asyncHandle -> do
     liftIO $ link asyncHandle
-    (Just batchConsumed) :: Maybe (EventStreamBatch (DataChangeEvent Foo)) <-
-      eventsProcessConduit (Just consumeParametersSingle) myEventTypeName Nothing headC
-    liftIO $ isJust (batchConsumed^.L.events) @=? True
+    subscriptionProcess (Just consumeParametersSingle) subscriptionId (storeBatch batchTv)
```
etorreborre: It's not so trivial to test a subscription... But I totally understand what you are doing.

tests/Network/Nakadi/Examples/Subscription/Process.hs
```diff
@@ -7,7 +7,7 @@ import           Data.Aeson
 import qualified Network.Nakadi as Nakadi
 import Network.Nakadi (MonadNakadi)
 import Control.Monad.Logger
-import Control.Monad.IO.Unlift
+import Control.Monad.Catch (MonadMask)
```
etorreborre: Why is this change necessary?

etorreborre:
It looks good to me. Apparently you are also cleaning some of the imports, or it is to make the compilation pass with updated dependencies?

tests/Network/Nakadi/Connection/Test.hs
```diff
@@ -14,6 +14,7 @@ import           Network.Wai.Handler.Warp
 import           System.IO.Unsafe
 import           Test.Tasty
 import           Test.Tasty.HUnit
+import           UnliftIO.Concurrent
```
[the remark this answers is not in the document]
mtesseract: Just checked, looks like it is for bringing 'threadDelay' into scope.

tests/Network/Nakadi/EventTypes/Test.hs
mtesseract: Yes... This test suite has grown in a more ad-hoc fashion. Would actually be nice to clean this up at some point and provide reasonable building blocks that allow for nicer test writing.

tests/Network/Nakadi/Examples/Subscription/Process.hs
mtesseract: These are actually two changes. Control.Monad.IO.Unlift was previously required for bringing MonadUnliftIO into scope, which looks like it is now done simply by importing the new ClassyPrelude.

On the other hand, MonadMask is required is required for Nakadi.subscriptionProcess and it seems like it is not exposed anymore by the previously imported modules.

mtesseract:
It was not really my intention to clean up imports, primarily it was simply directed by the compiler errors. :-)
(Not saying that this is not a good idea to be done!)

mtesseract:
Thanks for the review. If you are fine with this, then I will continue with merging.

etorreborre:
Yes, 👍

PR #94, opened 2018-03-09
Concurrent consumers

mtesseract:
Here it comes.
Subscriptions/Events.hs became to big, which is why I have decided to do some refactoring as well.


mtesseract:
At some point the subscription publishing/consumption parts of the test suite should be made less error-prone when it comes to timing issues in the test logic.

Feel free to give it a review, @etorreborre  :-D

src/Network/Nakadi/Internal/Committer.hs
```diff
@@ -0,0 +1,188 @@
+{-|
```
mtesseract: This is just refactoring. This is moved here from Subscription/Events.hs.

src/Network/Nakadi/Internal/Http.hs
```diff
@@ -72,17 +72,18 @@ import           Network.Nakadi.Internal.Util
 conduitDecode
   :: forall a b m
    . (FromJSON a, MonadNakadi b m)
-  => Config b -- ^ Configuration, containing the
-              -- deserialization-failure-callback.
-  -> ConduitM ByteString a m () -- ^ Conduit deserializing bytestrings
+  => ConduitM ByteString a m () -- ^ Conduit deserializing bytestrings
```
mtesseract: This is a minor simplification. Previously the caller extracted the nakadi config anyway, thus it was fine to simply pass it to `conduitDecode` as well. But now the caller doesn't need to do this and extracting the nakadi config can also simply be done in this function.

src/Network/Nakadi/Internal/Types.hs
```diff
@@ -170,6 +170,9 @@ instance (Monad b, MonadState s m) => MonadState s (NakadiT b m) where
   get = lift get
   put = lift . put
 
+instance (Monad b, MonadResource m) => MonadResource (NakadiT b m) where
```
mtesseract: The creation/destruction of worker threads requires ResourceT'ts allocate/dellocate guarantees.
We don't want to keep worker threads running if something goes wrong.
This means: when you consume a subscription, you need runResourceT around it.

src/Network/Nakadi/Internal/Worker.hs
```diff
@@ -0,0 +1,183 @@
...
+import           Network.Nakadi.Subscriptions.Subscription
+
+import           Network.Nakadi.Internal.Committer
+import qualified Network.Nakadi.Internal.Lenses            as L
+import           Network.Nakadi.Internal.Types
+import           UnliftIO.Async
+import           UnliftIO.STM
+
+-- | Data type denoting an asynchronous worker.
+data Worker a = Worker { _queue :: TBQueue (SubscriptionEventStreamBatch a)
```
mtesseract: A worker is essentially an async thread with a queue over which it receives batche.s

src/Network/Nakadi/Internal/Worker.hs
```diff
@@ -0,0 +1,183 @@
...
+                       , _async :: Async ()
+                       }
+
+-- | Data type containing a non-empty list of worker references.
+data WorkerRegistry a =
```
mtesseract: It is guaranteed that a WorkerRegistry always contains at least one worker.

src/Network/Nakadi/Subscriptions/Events.hs
```diff
@@ -127,187 +136,13 @@ subscriptionProcessHandler
...
-  catchAny (subscriptionCursorCommit eventStream [cursor]) $ \ exn -> nakadiLiftBase $
-    case config^.L.logFunc of
-      Just logFunc -> logFunc "nakadi-client" LevelWarn $ toLogStr $
-        "Failed to commit cursor " <> tshow cursor <> ": " <> tshow exn
-      Nothing ->
-        pure ()
+  let nWorkers = config^.L.worker.L.nThreads
+  eventStream    <- buildSubscriptionEventStream subscriptionId response
+  workerRegistry <- spawnWorkers subscriptionId eventStream nWorkers processor
+  race_ (workersWait workerRegistry) $ runConduit $
```
mtesseract: Returns if either some worker terminates or if the stream consumption terminates.

src/Network/Nakadi/Config.hs
```diff
@@ -53,6 +54,11 @@ defaultRetryPolicy = fullJitterBackoff 2 <> limitRetries 5
 defaultCommitStrategy :: CommitStrategy
 defaultCommitStrategy = CommitSync
 
+-- | Default worker configuration.
+defaultWorkerConfig :: WorkerConfig
+defaultWorkerConfig =
+  WorkerConfig { _nThreads      = 1 }
```
etorreborre: Why such a whitespace :-)?

src/Network/Nakadi/Internal/Prelude.hs
```diff
@@ -1,7 +1,7 @@
 {-|
 Module      : Network.Nakadi.Prelude
 Description : Nakadi Client Prelude (Internal)
-Copyright   : (c) Moritz Schulte 2017
```
etorreborre: Maybe it's time for a global search and replace :-): https://github.com/mtesseract/nakadi-client/search?utf8=%E2%9C%93&q=Schulte?

etorreborre:
I need more time to review the main logic after a good night' sleep. I'll try to do that tomorrow.

mtesseract:
Sure, looking forward to that! Let me know if something is unclear or should be documented better.

etorreborre:
Sorry, still no time... Maybe tomorrow :-).

src/Network/Nakadi/Internal/Committer.hs
```diff
@@ -0,0 +1,188 @@
...
+  cursorsMap <- liftIO . atomically $ newTVar HashMap.empty
+  withAsync (cursorConsumer cursorsMap) $ \ asyncCursorConsumer -> do
+    link asyncCursorConsumer
+    Timer.withAsyncTimer timerConf $ \ timer -> forever $ do
+      Timer.wait timer
+      commitAllCursors eventStream cursorsMap
+
+  where -- | The cursorsConsumer drains the cursors queue and adds each
+        -- cursor to the provided cursorsMap.
+        cursorConsumer cursorsMap = forever . liftIO . atomically $ do
```
etorreborre: It could be a good idea to put the `cursorsMap` type here (like `commitSmartBuffer`) or comment on what exactly the keys are.

src/Network/Nakadi/Internal/Committer.hs
```diff
@@ -0,0 +1,188 @@
...
+          loop
+
+-- | Implementation of the 'CommitTimeBuffer' strategy: We use an
+-- async timer for committing cursors at specified intervals.
+subscriptionCommitter (CommitTimeBuffer millis) eventStream queue = do
+  let timerConf = Timer.defaultConf
+                  & Timer.setInitDelay (fromIntegral millis)
+                  & Timer.setInterval  (fromIntegral millis)
+  cursorsMap <- liftIO . atomically $ newTVar HashMap.empty
+  withAsync (cursorConsumer cursorsMap) $ \ asyncCursorConsumer -> do
```
etorreborre: I am thinking that I would maybe extract this behaviour as its own "domain": there's a queue of cursors to commit and there should be a function consuming it periodically.

src/Network/Nakadi/Internal/Committer.hs
etorreborre: Also describe what is the `Int` in `HashMap PartitionName (Int, Cursor)`

src/Network/Nakadi/Internal/Committer.hs
```diff
@@ -0,0 +1,188 @@
...
+                                    & Timer.setInterval  (fromIntegral millisDefault)
+  cursorsMap <- liftIO . atomically $
+                newTVar (HashMap.empty :: HashMap PartitionName (Int, SubscriptionCursor))
+  withAsync (cursorConsumer cursorsMap) $ \ asyncCursorConsumer -> do
+    link asyncCursorConsumer
+    Timer.withAsyncTimer timerConf $ cursorCommitter cursorsMap nMaxEvents
+
+  where -- | The cursorsConsumer drains the cursors queue and adds
+        -- each cursor to the provided cursorsMap.
+        cursorConsumer cursorsMap = forever . liftIO . atomically $ do
```
etorreborre: This can possibly refactored with the previous code, where the only difference is `0` in one case and `nEvents` in the other case.

src/Network/Nakadi/Internal/Committer.hs
etorreborre: Also `0` seems to be irrelevant here so maybe we need another structure to hold the cursors to commit. I also have the impression that `cursorsMap` is passed to `commitAllCursors` which is not using this `Int` value. So `commitAllCursors` could be refactored to accept a "smaller" data structure (simply `HashMap PartitionName Cursor`)

src/Network/Nakadi/Internal/Committer.hs
```diff
@@ -0,0 +1,188 @@
...
+        -- least one such cursor is found.
+        maxEventsReached cursorsMap nMaxEvents = liftIO . atomically $ do
+          cursorsList <- HashMap.toList <$> readTVar cursorsMap
+          let cursorsCommit = filter (shouldBeCommitted nMaxEvents) cursorsList
+          if null cursorsCommit
+            then retry
+            else pure ()
+
+        -- | Returns True if the provided cursor should be committed.
+        shouldBeCommitted nMaxEvents cursor = cursor^._2._1  >= nMaxEvents
```
etorreborre: Maybe a small `CursorToCommit` data structure could be introduced to avoid `_2._1`:
```haskell
data CursorToCommit = 
  CursorToCommit
  { _nbOfEvents :: Int
  , _cursor :: Cursor
  }
```
Then 
```haskell
data CursorsToCommit = 
  CursorsToCommit 
  { _cursors :: HashMap PartitionName CursorToCommit
  }
```
I think the algorithm is already not trivial because it involves concurrency so it is probably to encapsulate the hashmaps with special purpose data structures.

src/Network/Nakadi/Internal/Committer.hs
```diff
@@ -0,0 +1,188 @@
...
+
+-- | This function commits all cursors in the provided cursorsMap.
+commitAllCursors
+  :: (MonadNakadi b m, MonadIO m)
+  => SubscriptionEventStream
+  -> TVar (HashMap PartitionName (Int, SubscriptionCursor))
+  -> m ()
+commitAllCursors eventStream cursorsMap = do
+  cursors <- liftIO . atomically $ swapTVar cursorsMap HashMap.empty
+  forM_ cursors $ \ (_nEvents, cursor) -> commitOneCursor eventStream cursor
```
etorreborre: I have been using `traverse` forever but I see more and more `for` in Haskell code so I'm getting used to it now :-).

src/Network/Nakadi/Internal/Worker.hs
```diff
@@ -0,0 +1,183 @@
...
+          runConduit $ producer .| mapM_C (sendToQueue commitQueue)
+
+  where sendToQueue commitQueue batch = liftIO . atomically $ do
+          let cursor  = batch^.L.cursor
+              events  = fromMaybe Vector.empty (batch^.L.events)
+              nEvents = length events
+          writeTBQueue commitQueue (nEvents, cursor)
+
+-- | Retrieve the 'PartitionIndexMap' for the given subscription.
+retrievePartitionIndexMap
```
etorreborre: I would call this `retrievePartitionMap` where there's no zipping with `[1..]` and add a function called `createWorkerRegistry` using that map and doing the zipping "internally".

src/Network/Nakadi/Internal/Worker.hs
```diff
@@ -0,0 +1,183 @@
...
+  :: (MonadIO m)
+  => WorkerRegistry a
+  -> ConduitM (SubscriptionEventStreamBatch a) Void m ()
+workerDispatchSink registry = awaitForever $ \ batch -> do
+  let worker = pickWorker (_workers registry) (_partitionIndexMap registry) batch
+  atomically $ writeTBQueue (_queue worker) batch
+
+-- | Given a 'SubscriptionEventStreamBatch', produce the worker that
+-- should handle the batch.
+pickWorker
```
etorreborre: I would hide the map here by passing the registry directly.

etorreborre:
Finally I got a bit of time to add some comments. I feel that some "encapsulation" would be nice to isolate a bit the committing logic from the structure / maps supporting it. Maybe this is my object-oriented experience speaking here but I think that would help in understanding the code.

mtesseract:
Thanks for your review, I will look at your comments as soon as possible.
By the way, on my work notebook I did some measurements and I was able to consume events at approx. 15700 E/s using 8 partitions / workers. This is, against a dockerized Nakadi running on localhost.
I am quite happy about this, as this is not event optimized yet.

src/Network/Nakadi/Internal/Worker.hs
mtesseract: Hmm. There is no map withing zipping with the indices here: These indices are the values of the map. This map is there for mapping pairs (event type name, partition name) to integer indices.

But maybe I misunderstood this comment.

src/Network/Nakadi/Internal/Worker.hs
mtesseract: Ok! Have changed it to be:

```
pickWorker
  :: WorkerRegistry a
  -> EventTypeName
  -> PartitionName
  -> Worker a
```

src/Network/Nakadi/Internal/Committer.hs
mtesseract: I have generalized this. Now there is a `StagedCursor a` type, where the `a` describes some enrichment for a cursor to be committed. In the smart case, `a` is `Int`. In the timer-buffer case, `a` is `()`.

Consequently, `commitAllCursors` is polymorphic in `a`.

src/Network/Nakadi/Internal/Committer.hs
mtesseract: As mentioned above, such a data type exists now. It is called `StagedCursor a`.

src/Network/Nakadi/Internal/Committer.hs
mtesseract: I am not yet sure what you have in mind here.
Will try to read this again later. If I still don't understand the remark, I will ask you to elaborate. :-)

mtesseract:
@etorreborre I have updated the PR.
But I think I still need to figure out what exactly you have in mind in terms of encapsulating.
Generally speaking, I like encapsulation, so I see value in that.
I cannot really comment on the OO background remark, as I don't really have an OO background. :-/

mtesseract:
I have now isolated the committing strategies.

src/Network/Nakadi/Internal/Committer/Shared.hs
```diff
@@ -0,0 +1,77 @@
...
+        pure ()
+
+-- | Implementation for the 'CommitNoBuffer' strategy: We simply read
+-- every cursor and commit it in order.
+unbufferedCommitLoop
+  :: (MonadNakadi b m, MonadIO m)
+  => SubscriptionEventStream
+  -> TBQueue (Int, SubscriptionCursor)
+  -> m ()
+unbufferedCommitLoop eventStream queue = do
```
etorreborre: Can we reuse the code from https://github.com/mtesseract/nakadi-client/pull/94/files/c4e02c10a22846a5bb0f5a5d69ee3f737477bb91..a3ba5c3cfb45bcb9c8a0d1e3efcf9c2f4ab72996#diff-44e998763ba0d9d12f274099c37c53f5R35?

etorreborre:
Hi @mtesseract I haven't the time for an in-depth review but I scrolled through the changes and the extraction of the various components looks good to me. I just left a comment about a refactoring opportunity (I think).

src/Network/Nakadi/Internal/Committer/Shared.hs
mtesseract: Will look into that later. thanks!

mtesseract:
@etorreborre, I have reduced the code duplication that you have spotted. :-)
Would you then be fine with merging or would you feel uncomfortable with it because of lack of time?

src/Network/Nakadi/Internal/Committer/Shared.hs
mtesseract: done

src/Network/Nakadi/Config.hs
```diff
@@ -166,6 +173,9 @@ setCommitStrategy
   -> Config m
 setCommitStrategy = (L.commitStrategy .~)
 
+setWorkerThreads :: Int -> Config m -> Config m
```
erthalion: I wonder why other functions have some commentaries, but not this one? I mean, it's obvious, what it does, as well as it obvious for others (e.g. setBatchLimit), but looks a bit inconsistent.

src/Network/Nakadi/Internal/Committer/SmartBuffer.hs
```diff
@@ -0,0 +1,144 @@
...
+maxEventsReached :: MonadIO m => TVar StagedCursorsWithCounter -> Int -> m ()
+maxEventsReached stagedCursorsTv nMaxEvents = liftIO . atomically $ do
+  stagedCursors <- readTVar stagedCursorsTv
+  let cursorsList   = HashMap.elems (stagedCursors^.L.cursorsMap)
+      cursorsCommit = filter (shouldBeCommitted nMaxEvents) cursorsList
+  if null cursorsCommit
+    then retry
+    else pure ()
+
+-- | Returns True if the provided staged cursor with should be
```
erthalion: `staged cursor with should be`, something is missing after `with`?

src/Network/Nakadi/Internal/Committer/TimeBuffer.hs
```diff
@@ -0,0 +1,63 @@
...
+committerTimeBuffer
+  :: ( MonadNakadi b m
+     , MonadUnliftIO m
+     , MonadMask m )
+  => Int32
+  -> SubscriptionEventStream
+  -> TBQueue (Int, SubscriptionCursor)
+  -> m ()
+committerTimeBuffer millis eventStream queue = do
+  let timerConf = Timer.defaultConf
```
erthalion: Is there a possibility to reuse this code with the corresponding part of smart buffer implementation?

src/Network/Nakadi/Internal/Types/Committer.hs
```diff
@@ -0,0 +1,15 @@
...
+
+{-# LANGUAGE StrictData            #-}
+
+import           Data.HashMap.Strict                   (HashMap)
+
+import           Network.Nakadi.Internal.Types.Service
+
+data StagedCursor a =
+  StagedCursor { _cursor     :: SubscriptionCursor
+               , _enrichment :: a }
```
erthalion: So for now `enrichment` contains only a number of events that happened between two timer calls? If there are no immediate plans to add something more, I would suggest to call it differently, because in the underlying code it's confusing to associate `enrichment` with nMaxEvents

src/Network/Nakadi/Config.hs
mtesseract: done

src/Network/Nakadi/Internal/Committer/SmartBuffer.hs
mtesseract: fixed

src/Network/Nakadi/Internal/Committer/TimeBuffer.hs
mtesseract: will look into that later.

mtesseract:
Any update on this?

etorreborre:
No new review from me, nor the time to try it on our current setup. I will try to give it a go when I load test our next service consuming events from SPP because this might add a good boost in performance when we will bootstrap the service.

src/Network/Nakadi/Internal/Types/Committer.hs
```diff
@@ -0,0 +1,15 @@
+module Network.Nakadi.Internal.Types.Committer where
+
+{-# LANGUAGE StrictData            #-}
```
erthalion: Is it even correct to add a language pragma after a module definition?

mtesseract:
I will open a new PR for this.

PR #97, opened 2018-04-18
Upgraded to lts-11.5

etorreborre:
@mtesseract I'm giving a go to a more recent version of the lts (aiming for nakadi-client 0.6.0.0) so that we can align all our marbles.


mtesseract:
Cool, thanks!

PR #98, opened 2018-06-14
Support time lag retrieval for subscription stats

mtesseract:
This fixes the per-subscription statistics retrieval and implements support for the new `show_time_lag` parameter of the Nakadi API.

The statistics retrieval never worked before, thus I felt free to improve the general API in this regard.

In particular I have renamed the type `SubscriptionEventTypeStatsResult` to `SubscriptionStats` — the Nakadi Swagger does not contain a specific name for this object type.

There is some reformatting being done by brittany, sorry for the noise.
I have tested the API call manually, but a test should be added.

I wouldn't mind merging this right away, though.

### DIFF
--- a/src/Network/Nakadi/Config.hs
+++ b/src/Network/Nakadi/Config.hs
@@ -31,18 +31,22 @@ module Network.Nakadi.Config
   , setFlowId
   , setCommitStrategy
   , defaultConsumeParameters
-  ) where
+  , setShowTimeLag
+  )
+where
 
 import           Network.Nakadi.Internal.Prelude
 
 import           Control.Lens
 import           Control.Retry
-import           Network.HTTP.Client                   (Manager,
-                                                        ManagerSettings)
-import           Network.HTTP.Client.TLS               (newTlsManagerWith)
+import           Network.HTTP.Client            ( Manager
+                                                , ManagerSettings
+                                                )
+import           Network.HTTP.Client.TLS        ( newTlsManagerWith )
 import           Network.Nakadi.Internal.Config
 import           Network.Nakadi.Internal.HttpBackendIO
-import qualified Network.Nakadi.Internal.Lenses        as L
+import qualified Network.Nakadi.Internal.Lenses
+                                               as L
 import           Network.Nakadi.Internal.Types
 
 -- | Default retry policy.
@@ -60,20 +64,21 @@ newConfig
   => HttpBackend b
   -> Request           -- ^ Request Template
   -> Config b          -- ^ Resulting Configuration
-newConfig httpBackend request =
-  Config { _consumeParameters              = Nothing
-         , _manager                        = Nothing
-         , _requestTemplate                = request
-         , _requestModifier                = pure
-         , _deserializationFailureCallback = Nothing
-         , _streamConnectCallback          = Nothing
-         , _logFunc                        = Nothing
-         , _retryPolicy                    = defaultRetryPolicy
-         , _http                           = httpBackend
-         , _httpErrorCallback              = Nothing
-         , _flowId                         = Nothing
-         , _commitStrategy                 = defaultCommitStrategy
-         }
+newConfig httpBackend request = Config
+  { _consumeParameters              = Nothing
+  , _manager                        = Nothing
+  , _requestTemplate                = request
+  , _requestModifier                = pure
+  , _deserializationFailureCallback = Nothing
+  , _streamConnectCallback          = Nothing
+  , _logFunc                        = Nothing
+  , _retryPolicy                    = defaultRetryPolicy
+  , _http                           = httpBackend
+  , _httpErrorCallback              = Nothing
+  , _flowId                         = Nothing
+  , _commitStrategy                 = defaultCommitStrategy
+  , _subscriptionStats              = Nothing
+  }
 
 -- | Producs a new configuration, with mandatory HTTP manager, default
 -- consumption parameters and HTTP request template.
@@ -84,8 +89,8 @@ newConfigIO = newConfig httpBackendIO
 
 -- | Produce a new configuration, with optional HTTP manager settings
 -- and mandatory HTTP request template.
-newConfigWithDedicatedManager ::
-  (MonadIO b, MonadMask b, MonadIO m)
+newConfigWithDedicatedManager
+  :: (MonadIO b, MonadMask b, MonadIO m)
   => ManagerSettings -- ^ Optional 'ManagerSettings'
   -> Request         -- ^ Request template for Nakadi requests
   -> m (Config b)    -- ^ Resulting Configuration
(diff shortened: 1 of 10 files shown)


src/Network/Nakadi/Subscriptions/Stats.hs
```diff
@@ -18,39 +18,52 @@ API.
...
+import           Network.Nakadi.Internal.Util
 
 path :: SubscriptionId -> ByteString
 path subscriptionId =
-  "/subscriptions/"
-  <> subscriptionIdToByteString subscriptionId
-  <> "/cursors"
+  "/subscriptions/" <> subscriptionIdToByteString subscriptionId <> "/stats"
 
 -- | @GET@ to @\/subscriptions\/SUBSCRIPTION\/cursors@. Low level
```
etorreborre: This comment needs to be updated with `/stats` instead of `/cursors`. Same thing for the other function below.

src/Network/Nakadi/Subscriptions/Stats.hs
```diff
@@ -18,39 +18,52 @@ API.
...
 
 -- | @GET@ to @\/subscriptions\/SUBSCRIPTION\/cursors@. High level
 -- interface for Subscription Statistics retrieval.
-subscriptionStats ::
-  MonadNakadi b m
+subscriptionStats
+  :: MonadNakadi b m
   => SubscriptionId                        -- ^ Subscription ID
+  -> [SubscriptionStatsParameter]
   -> m (Map EventTypeName [PartitionStat]) -- ^ Subscription
```
etorreborre: I am generally not satisfied with returning `Map` as a datatype from an API because it feels like exposing implementation details. `SubscriptionStats` seems to be fine because it maps 1-to-1 with the JSON returned by Nakadi but maybe I would reserve this name for a wrapper around the `Map` and would find another name for the "low-level".

etorreborre:
👍

src/Network/Nakadi/Subscriptions/Stats.hs
mtesseract: Done

src/Network/Nakadi/Subscriptions/Stats.hs
mtesseract: I'm not sure I understand the criticism regarding `Map`. But maybe we can discuss possible API changes related to this question in a separate issue / PR and restrict this one to implement the show_time_lag functionality?

What about this approach of providing parameter flags as a list? Does this seem reasonable?

Thanks for your review.

etorreborre: My experience with providing maps in API hasn't been so good but possibly because the underlying domain was changing too much, using different keys or having additional keyed values. In the case of Nakadi I suppose the API is kind of stable or not so changing that we need to encapsulate every map.

I guess the drawbacks with parameters as a list is that you make mistakes by having more than one in the list. Maybe one alternative, which would work because there aren't too many of them, would be to use the typelevel-sets project?  If there are too many parameters the compiler becomes dog-slow :-(.

mtesseract: I also assume that we can safely assume that the return value of `subscriptionStats` yields a map keyed by `EventTypeName` and providing (possibly many) per-partition statistics. If this changes, I guess we are talking about a different / new endpoint.
Translating the value from some opaque `SubscriptionStats` to some more concrete `Map` removes — in my opinion — unnecessary unwrapping / encapsulattion.

Not sure if I correctly understand your argument regarding parameters as a list. I *think* you are referring to the situation of mutually exclusive parameters and the possibility to reject programs which are invalid in this regard, is this correct? i.e., nobody would prevent me from providing `[ShowTimeLag, WithoutTimeLag]`? Personally I have no experience with type-level-sets. But before going type level I would find it very important to have a clear picture of the pros/cons situation.

Don't get me wrong, I am really not sure if this parameters-as-list approach is the most reasonably style to implement here. Especially since we should find a way that does not only satisfy the requirements for a single endpoint, but might be a more general approach for configuring specific Nakadi requests.

Alternatively, we could also change the function such that it reads:
```
subscriptionStats :: SubscriptionStatsConf -> SubscriptionId -> ...
```

where the constructors for `SubscriptionStatsConf` are not exported, instead of export some smart constructors:

```
subscriptionStatsConfWithTimeLag :: SubscriptionStatsConf
subscriptionStatsConfPlain :: SubscriptionStatsConf
```

In terms of extendability, this would allow us to provide additional smart constructors in the future, if the API allows even more configurability.

mtesseract: Or we just take the easy way out: We already have a configuration object, it is implicitly contained in the Nakadi monad.

How about we simply implement:

```
setShowTimeLag :: Bool -> Config m -> Config m
```
Then, subscriptionStats can remain simple:

```
subscriptionStats :: MonadNakadi b m => SubscriptionId -> m (Map EventTypeName [PartitionStat])
```

And users can also it as follows:

```
local (setShowTimeLag %~ True) $ subscriptionStats ...
```

I think I would prefer this way!

etorreborre: That seems reasonable indeed :-)

tests/Network/Nakadi/Subscriptions/Test.hs
```diff
@@ -12,81 +12,85 @@ import           ClassyPrelude
...
+  "Subscriptions"
+  [ testSubscriptionsStats conf
+  , testSubscriptionsProcessing conf
+  , testCase "SubscriptionsList"               (testSubscriptionsList conf)
+  , testCase "SubscriptionsCreateDelete"       (testSubscriptionsCreateDelete conf)
   , testCase "SubscriptionDoubleDeleteFailure" (testSubscriptionsDoubleDeleteFailure conf)
   ]
 
 testSubscriptionsList :: Config App -> Assertion
-testSubscriptionsList conf = runApp . runNakadiT conf $ do
```
mtesseract: I did this change in order to reduce formatting noise by brittany. Otherwise the whole function would have been reindented.

src/Network/Nakadi/Config.hs
```diff
@@ -60,20 +64,21 @@ newConfig
   => HttpBackend b
   -> Request           -- ^ Request Template
   -> Config b          -- ^ Resulting Configuration
-newConfig httpBackend request =
```
mtesseract: Formatting noise, only `_subscriptionStats` has been added.

src/Network/Nakadi/Config.hs
```diff
@@ -84,8 +89,8 @@ newConfigIO = newConfig httpBackendIO
 
 -- | Produce a new configuration, with optional HTTP manager settings
 -- and mandatory HTTP request template.
```
mtesseract: Most of this is formatting noise. Only `setShowTimeLag` is new.

src/Network/Nakadi/Internal/Util.hs
```diff
@@ -18,38 +18,39 @@ module Network.Nakadi.Internal.Util
   , decodeThrow
   , sequenceSnd
   , extractQueryParametersFromPath
```
mtesseract: With the exception of `encodeStrict` everything is formatting noise here.

tests/Network/Nakadi/Subscriptions/Test.hs
```diff
@@ -12,81 +12,85 @@ import           ClassyPrelude
...
-import qualified Network.Nakadi.Lenses       as L
+import qualified Network.Nakadi.Lenses         as L
 import           Network.Nakadi.Tests.Common
 import           Test.Tasty
 import           Test.Tasty.HUnit
-import Network.Nakadi.Subscriptions.Processing.Test
+import           Network.Nakadi.Subscriptions.Processing.Test
+import           Network.Nakadi.Subscriptions.Stats.Test
 
 testSubscriptions :: Config App -> TestTree
```
mtesseract: More noise, I have simply added the new tests in this module.

mtesseract:
As discussed, I have added the `setShowTimeLag` function and simplified the `subscriptionStats` API.

When I compared our model for `PartitionStat` with the Nakadi API (https://nakadi.io/manual.html#definition_SubscriptionEventTypeStats), I have noticed that we wrongly regarded `stream_id` and `unconsumed_events` as mandatory. I have fixed this.

I am currently trying out the Haskell IDE Engine (HIE), it comes bundled with brittany integration (https://github.com/lspitzner/brittany), which has sadly added quite some formatting noise, sorry for that.

I have also added missing tests for subscriptionStats.

Ok to merge?

tests/Network/Nakadi/Subscriptions/Stats/Test.hs
```diff
@@ -0,0 +1,84 @@
...
+testSubscriptionStatsWithTimeLag :: Config App -> Assertion
+testSubscriptionStatsWithTimeLag conf = do
+  stats <- produceSubscriptionStats (setShowTimeLag True conf)
+  let withoutTimeLags =
+        stats
+          & toList
+          & concat
+          & map (^. L.consumerLagSeconds)
+          & filter isNothing
+  liftIO $ [] @=? withoutTimeLags
```
etorreborre: Can we also show that there is at least one time lag in the list, because it might as well be empty?

etorreborre:
👍 ok to merge but I think that one test could be improved.

etorreborre:
👍

PR #104, opened 2018-06-20
Added typeclasses for business and datachange events

### LINKED ISSUE #58: Add BusinessEvent type support
Hi @mtesseract, I'm trying to add BusinessEvent and got a question. 
For DataChangeEvent type we have an abstract parameter "a" for payload. Trying to do the same for BusinessEvent I found some difficulties. From the definition of business events https://zalando.github.io/nakadi/manual.html#definition_BusinessEvent we need to mixin "metadata"   field into our payload. 
Trying to write aeson serializer I got something like this: 
```
data BusinessEvent a = BusinessEvent
  { _payload  :: a
  , _metadata :: Metadata
  } deriving (Eq, Show, Generic)

instance FromJSON a => FromJSON (BusinessEvent a) where
  parseJSON = withObject "BusinessEvent" $ \obj -> BusinessEvent
    <$> parseJSON (Object obj)
    <*> obj .: "metadata"

instance ToJSON a => ToJSON (BusinessEvent a) where
  toJSON (BusinessEvent payload metadata) = case toJSON payload of
    Object obj -> Object $ obj `union` fromList ["metadata" .= toJSON metadata]
    other -> other
```
The problem is in ToJSON instance. Type parameter "a" doesn't mean that it will be serialized into aeson Object. It could also be the other Value which makes impossible to mixin metadata into it. 
How do you think, what the better way to define BusinessEvent serializer?

vitold:
@mtesseract Hi, that's my suggestion for the event models. WIP

Closes #58

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -48,6 +48,7 @@ dependencies:
 - monad-logger
 - text
 - lens
+- lens-aeson
 - http-types
 - http-client
 - http-client-tls
--- a/src/Network/Nakadi/EventTypes/Events.hs
+++ b/src/Network/Nakadi/EventTypes/Events.hs
@@ -17,6 +17,7 @@ This module implements the
 {-# LANGUAGE ScopedTypeVariables   #-}
 {-# LANGUAGE TupleSections         #-}
 {-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE TypeApplications       #-}
 
 module Network.Nakadi.EventTypes.Events
   ( eventsPublish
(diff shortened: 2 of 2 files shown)


mtesseract:
Hi, thanks for the PR.

I'm not sure I understand this, though.

In my opinion, we should have less law-less typeclasses instead of more. It is not obvious to me why having support for Business Events requires an implementation that seems so different than the rest of the event handling code.

This PR introduces four more typeclasses — that's a lot.
And why do we need seperate publishing functions (`publishBusinessEvents`) for these events?

Some kind of rationale for this is missing for me. On the other hand, I haven't looked at Business Events for quite some time. So, I will change this if time permits.

Thanks!

mtesseract:
I propose the following PR instead: https://github.com/mtesseract/nakadi-client/pull/106/files

mtesseract:
Hi,
thanks again but I have closed this in favour of #106.

PR #106, opened 2018-06-23
Implement BusinessEvents

mtesseract:
Implement BusinessEvents with custom JSON deserialization/serialization implementations.
Including tests.

### DIFF
--- a/src/Network/Nakadi/Internal/Types/Service.hs
+++ b/src/Network/Nakadi/Internal/Types/Service.hs
@@ -16,6 +16,7 @@ Types modelling the Nakadi Service API.
 {-# LANGUAGE LambdaCase            #-}
 {-# LANGUAGE StrictData            #-}
 {-# LANGUAGE TemplateHaskell       #-}
+{-# LANGUAGE RecordWildCards #-}
 
 module Network.Nakadi.Internal.Types.Service where
 
@@ -32,6 +33,7 @@ import           Data.Time.ISO8601
 import           Data.UUID
 import           Data.Vector                    ( Vector )
 import           GHC.Generics
+import qualified Data.HashMap.Strict           as HashMap
 
 import           Network.Nakadi.Internal.Json
 import           Network.Nakadi.Internal.Types.Util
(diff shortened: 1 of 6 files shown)


mtesseract:
@etorreborre @erthalion
Care to review this?
This should finally bring support for Business Events to nakadi-client.
There is another PR open for this (https://github.com/mtesseract/nakadi-client/pull/104), but I don't consider this be a consistent addition to nakadi-client.

erthalion:
Sure, will jump on it this evening

src/Network/Nakadi/Internal/Types/Service.hs
```diff
@@ -785,3 +787,56 @@ data DataChangeEventEnriched a = DataChangeEventEnriched
...
+-- | Type modelling a "Business Event". Their JSON encodings are special since the payload
+-- object is directly enriched with a @metadata@ field. "Data Change Events" on the other side
+-- are JSON-encoded such that the complete event payload is contained in a seperate object field.
+--
+-- On the Haskell API side we split payload from meta data, which requires us to write custom
+-- 'ToJSON' and 'FromJSON' implementations.
+
+data BusinessEvent a = BusinessEvent
+  { _payload  :: a             -- ^ Event Payload.
+  , _metadata :: EventMetadata -- ^ Even Metadata.
```
erthalion: typo, `Even` -> `Event` ?

src/Network/Nakadi/Internal/Types/Service.hs
```diff
@@ -785,3 +787,56 @@ data DataChangeEventEnriched a = DataChangeEventEnriched
...
+        -- the object with metadata.
+        Object (o <> metadata)
+      _ ->
+        -- Otherwise, produce an object that only contains metadata.
+        Object metadata
+    where metadata = HashMap.fromList [("metadata", toJSON _metadata)]
+
+-- | Type modelling a Nakadi-enriched "Business Event". JSON encoding is basically the same as
+-- for the non-enriched Business Events.
+data BusinessEventEnriched a = BusinessEventEnriched
```
erthalion: I'm a bit confused about how similar `BusinessEvent` and `BusinessEventEnrivhed` are. Does it makes sense to use one `BusinessEvent` tyoe, but introduce something like:
```haskell
data EventMeta = EventMetaShort EventMetadata | EventMetaFull EventMetadataEnriched
```
and then:
```haskell
data BusinessEvent a = BusinessEvent
{ _payload   :: a
, _metadata :: EventMeta
}
```
What do you think?

tests/Network/Nakadi/EventTypes/BusinessEvents/Test.hs
```diff
@@ -0,0 +1,182 @@
...
+    }
+  , _partitionKeyFields   = Nothing
+  , _defaultStatistic     = Nothing
+  , _options              = Nothing
+  }
+
+eventSpecA :: EventSpec (BusinessEvent TestEventA)
+eventSpecA = EventSpec genBusinessEventA eventTypeA
+
+genBusinessEventA :: IO (BusinessEvent TestEventA)
```
erthalion: What about tests for `BusinessEventEnriched`?

src/Network/Nakadi/Internal/Types/Service.hs
mtesseract: I tend to think that this is not a good idea.
Metadata and Nakadi-enriched Metadata are currently differentiable at the type-level, not just two different constructors for the same type. I think it's good so stay like this: It reduces pattern matching and these are — in my opinion — really two different use-cases of metadata with different mandatory fields. (Question: How would we write the JSON parsers for this, when would it be clear that it is EventMataShort and when that it is EventMetaFull that we are deserializing? When would we know when mandatory fields are missing?)

On the other hand, if we were about to touch this, then this should be part of a different PR. Right now, it is completely in sync with DataChangeEvent.

tests/Network/Nakadi/EventTypes/BusinessEvents/Test.hs
mtesseract: Yeah, actually the test should include both types....

You are right. Will add this.

src/Network/Nakadi/Internal/Types/Service.hs
mtesseract: Fixed

etorreborre:
I have nothing to add on top of what @erthalion reviewed so far. So if all comments are addressed :+1:

src/Network/Nakadi/Internal/Types/Service.hs
erthalion: Ok, makes sense

tests/Network/Nakadi/EventTypes/BusinessEvents/Test.hs
mtesseract: I have extended the test. The type EventSpec now requires three type parameters: a, b and c. `a` denotes the non-enriched event type used for publishing, `b` denotes the enriched event type used for consuming and `c` denotes the payload type included in both `a` and `b`. EventSpec then contains getter functions of the form `a → c` and `b → c`. We need this for comparing that what has been consumed is exactly what has been published in terms of payload.

I tried to write the test non-invasive completely in its own module. At some point we can consider to refactor this and use some of the techniques used in the new test file also in other tests.

erthalion: Cool, :+1:

PR #109, opened 2018-06-26
Separate subscription types

### LINKED ISSUE #103: Use seperate data types for Subscriptions-to-be-created and Subscriptions-already-created
Currently these two cases are handled by the same data type "Subscription". It's field "id" (a Subscription ID) is optional. Which is, of course, a poor attempt at covering these two different use-cases.

When creating a subscription the field "id" should not be provided, not even `Nothing`. When an existing subscription is returned it should contained a mandatory subscription ID, not a `Maybe`.

It probably makes sense to come with two distinct data type names and use them accordingly.

mtesseract:
closes #103

Introduces a new type `SubscriptionRequest`.
This reorganizes and simplifies the type `Subscription`, most importantly the Subscription ID becomes **mandatory**.

Also, I have simplified (for the user) the handling of the subscription position in these datatypes.
In JSON the subcription position is given by two distinct object fields: read_from and cursors. On the sid of the Haskell API this is now merged in a new ADT SubscriptionPosition. This is a much better modelling of the Nakadi API in my opinion.

Furthermore, a newtype has been added: `ConsumerGroup`.

### DIFF
--- a/src/Network/Nakadi/Internal/Lenses.hs
+++ b/src/Network/Nakadi/Internal/Lenses.hs
@@ -1,7 +1,7 @@
 {-|
 Module      : Network.Nakadi.Lenses
 Description : Nakadi Client Library Lenses (Internal)
-Copyright   : (c) Moritz Schulte 2017, 2018
+Copyright   : (c) Moritz Clasmeier 2017, 2018
 License     : BSD3
 Maintainer  : mtesseract@silverratio.net
 Stability   : experimental
@@ -59,6 +59,7 @@ makeNakadiLenses ''SubscriptionCursor
 makeNakadiLenses ''CursorCommit
 makeNakadiLenses ''SubscriptionsListResponse
 makeNakadiLenses ''Subscription
+makeNakadiLenses ''SubscriptionRequest
 
 instance HasNakadiId StreamId Text where
   id f (StreamId a) = StreamId <$> f a
--- a/tests/Network/Nakadi/Internal/Types/Test.hs
+++ b/tests/Network/Nakadi/Internal/Types/Test.hs
@@ -67,11 +67,11 @@ testDecodeCursorCommitResults = do
   let eventType = "http4s-nakadi.test-event"
       offset =  "001-0001-000000000000007598"
       cursorToken = "3bb3a590-ede5-43a9-981e-2bea26347c99"
-      partition = "0"
+      partitionName = "0"
       result = CursorCommitResultOutdated
       cursorCommitResultsJson
        = encode [aesonQQ|
-          { items: [ { cursor: { partition: #{partition}
+          { items: [ { cursor: { partition: #{partitionName}
                                , offset: #{offset}
                                , event_type: #{eventType}
                                , cursor_token: #{cursorToken}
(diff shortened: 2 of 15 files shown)


src/Network/Nakadi/Internal/Lenses.hs
```diff
@@ -1,7 +1,7 @@
 {-|
 Module      : Network.Nakadi.Lenses
 Description : Nakadi Client Library Lenses (Internal)
-Copyright   : (c) Moritz Schulte 2017, 2018
+Copyright   : (c) Moritz Clasmeier 2017, 2018
```
etorreborre: Maybe it's time for a PR doing the full search and replace :-)?

tests/Network/Nakadi/Internal/Types/Test.hs
```diff
@@ -2,22 +2,80 @@
...
+testDecodeSubscriptionRequest = do
+  let
+    req pos = SubscriptionRequest
+      { _owningApplication    = "test"
+      , _eventTypes           = ["event1"]
+      , _consumerGroup        = Nothing
+      , _subscriptionPosition = pos
+      }
+    req'End
+      = "\
```
etorreborre: Would it be a good idea to use a quasi quoting library for json to improve readability here and to get compilation errors if the json is incorrect in the tests?

tests/Network/Nakadi/Internal/Types/Test.hs
```diff
@@ -2,22 +2,80 @@
...
+         \{ \"owning_application\": \"test\" \
+         \, \"event_types\": [\"event1\"] \
+         \, \"read_from\": \"end\" \
+         \}"
+    req'Begin
```
etorreborre: I never used quotes inside identifiers, I didn't realize this was possible.

tests/Network/Nakadi/Subscriptions/Test.hs
```diff
@@ -52,43 +53,37 @@ testSubscriptionsList conf = run $ do
...
   }
 
 testSubscriptionsCreateDelete :: Config App -> Assertion
 testSubscriptionsCreateDelete conf = runApp . runNakadiT conf $ do
   recreateEvent myEventType
   subscription <- subscriptionCreate (mySubscription Nothing)
-  liftIO $ True @=? isJust (subscription ^. L.id)
-  let (Just subscriptionId) = subscription ^. L.id
-  subscriptionDelete subscriptionId
+  subscriptionDelete (subscription ^. L.id)
```
etorreborre: Much better!

etorreborre:
I agree that's a much better modelling of the endpoint. 👍

tests/Network/Nakadi/Internal/Types/Test.hs
mtesseract: You mean something like aeson-qq, right? Never used it before! Will check it out.

src/Network/Nakadi/Internal/Lenses.hs
mtesseract: Might be a good idea. :)

mtesseract:
Thank you very much for the review!
Will check the quasi quoting before merging.

mtesseract:
Fine to merge, right?

etorreborre:
Beautiful.

PR #113, opened 2018-07-01
Event source

mtesseract:
Implement simply event sourcing.

Up for discussion, probably requires some improvements.

This PR also removes the weirdness of the previous subscriptionProcess API where a ConsumeParameters value had to be provided. This is now implicit in the main configuration.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -69,6 +69,7 @@ dependencies:
 - stm
 - unliftio >= 0.2.4.0 && < 0.3
 - async-timer >= 0.2.0.0 && < 0.3
+- stm-chans
 library:
   source-dirs: src
   exposed-modules:
--- a/src/Network/Nakadi/Config.hs
+++ b/src/Network/Nakadi/Config.hs
@@ -147,9 +147,6 @@ setFlowId flowId = L.flowId .~ Just flowId
 setCommitStrategy :: CommitStrategy -> Config m -> Config m
 setCommitStrategy = (L.commitStrategy .~)
 
-setShowTimeLag :: Bool -> Config m -> Config m
-setShowTimeLag flag = L.subscriptionStats ?~ SubscriptionStatsConf {_showTimeLag = flag}
-
 -- | Set maximum number of uncommitted events.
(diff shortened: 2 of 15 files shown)


mtesseract:
@etorreborre I am still thinking about some aspects here, but this might be a start.

package.yaml
```diff
@@ -69,6 +69,7 @@ dependencies:
 - stm
 - unliftio >= 0.2.4.0 && < 0.3
 - async-timer >= 0.2.0.0 && < 0.3
+- stm-chans
```
reactormonk: Some claim this is slow: https://hackage.haskell.org/package/unagi-chan

mtesseract: Thanks for your feedback!

To be honest, I've never really looked into the details of these benchmarks. But given that the different APIs are mostly easily substitutable, I personally would suggest to not open the box of aggressive optimization as part of this PR.

To do this right, we should also implement according benchmarking tests so that we can measure optimization benefits — something I would also prefer to keep seperate from this issue.

How about you open a general issue "Review usage/choice of communication channels" or something along these lines? This would then apply not only to this specific use of TBMQueue, but also more generally to other queue usages within nakadi-client.

mtesseract:
Hi,

just wanna add that I thought again about the semantics of the `streamLimit` setting in the context of `subscriptionSource` and I think it is reasonable as I it is right now in the PR:

What the PR basically does is shifting all responsibility for the `streamLimit` handling over to the Nakadi server. There's always the problem of Nakadi reconnects during streaming and doing the book keeping on the client side such that **exactly** `streamLimit` (distinct!) events are produced by the source is rather tricky.

As a side note, maybe, at some point, Nakadi would also implement a behaviour such that there are no reconnects if you are doing finite streaming (i.e., if `streamLimit` is specified).

All that being said, I think we should not put too much effort into these questions. What this PR tries to implement is a simple and deterministic behaviour where most actual work is delegated to the Nakadi server.

Feel free to do the review!

etorreborre:
That looks good to me, I will try this version with our nakadi-cli tool.

PR #118, opened 2018-07-12
Implement Concurrent Workers

mtesseract:
Let's discuss this new updated PR.


mtesseract:
```

                                          +----------------------------------------+
                                          |                                        |
                                          |   Subscription Consumption             |
                                          |                                        |
                                          |     Subscription/Events.hs             |
                                          |                                        |
                                          |                                        |
                                          |  Function: subscriptionProcessHandler  |
                                          |                                        |
                                          +----------------------------------------+
                                                             |
                                                             |
                                                             |
                                       +---------------------v--------------------------+
                                       |                                                |
                                       |       Worker Access Layer                      |
                                       |                                                |
                                       |       Internal/Worker.hs                       |
                                       |                                                |
                                       |    Spawns workers, creating a Worker Registry. |
                                       |    Dispatches batches to workers.              |
                                       |    Waits for workers to terminate.             |
                                       |                                                |
                                       |                                                |
                                       +------------------------------------------------+
                                                              |
                          +------------------------------------------------------------------------+
                          |                                                                        |
                          |                                                                        |
                          |                                                                        |
                          |                                                                        |
        +-----------------v-----------------+                                    +-----------------v-----------------+
        |                                   |                                    |                                   |
        |   Worker 1                        |                                    |   Worker n                        |
        |                                   |                                    |                                   |
        |     Internal/Worker.hs            |                                    |     Internal/Worker.hs            |
        |                                   |                                    |                                   |
        |     Function: subscriptionWorker  |                                    |     Function: subscriptionWorker  |
        |                                   |                                    |                                   |
        +-----------------------------------+                                    +-----------------------------------+
                         |                                   . . . . . .                           |
                         |                                                                         |
    +--------------------v-----------------------+                          +----------------------v---------------------+
    |                                            |                          |                                            |
    |  Committer                                 |                          |  Committer                                 |
    |                                            |                          |                                            |
    |  Internal/Committer.hs                     |                          |  Internal/Committer.hs                     |
    |                                            |                          |                                            |
    |  Functions:                                |                          |  Functions:                                |
    |                                            |                          |                                            |
    |  subscriptionSink (synchronous case)       |                          |  subscriptionSink (synchronous case)       |
    |  subscriptionCommitter (asynchronous case) |                          |  subscriptionCommitter (asynchronous case) |
    |                                            |                          |                                            |
    +--------------------------------------------+                          +--------------------------------------------+

```

reactormonk:
Maybe a bit late, but possibly relevant: https://github.com/roman/Haskell-capataz

mtesseract:
Have added some more code comments in the PR.

tests/Network/Nakadi/Examples/Subscription/Test.hs
```diff
@@ -42,10 +42,12 @@ testConsumption config = Nakadi.runNakadiT config $ do
...
-      events <- genEvents
-      void . async $ delayedPublish Nothing events
-      withAsync (dumpSubscription subscriptionId)
-        $ \_dumpHandle -> threadDelay (2 * 10 ^ 6) -- Give Nakadi some time to transmit the published events
+      events          <- genEvents
+      publisherHandle <- async $ delayedPublish Nothing events
+      link publisherHandle
+      withAsync (dumpSubscription subscriptionId) $ \dumpHandle -> do
+        link dumpHandle
+        threadDelay (5 * 10 ^ 6) -- Give Nakadi some time to transmit the published events
```
erthalion: Just out of curiosity, why did you increase this delay from 2 * 10^6 to 5 * 10^6 ?

mtesseract: This is a consequence of my local testing.

It seemed that my local Nakadi instance required some more startup time when partitioning was enabled. But maybe this can also be fine tuned with the available request parameters. I am not sure.

src/Network/Nakadi/Internal/Committer/SmartBuffer.hs
```diff
@@ -0,0 +1,132 @@
...
+import qualified Network.Nakadi.Internal.Lenses
+                                               as L
+import           Network.Nakadi.Internal.Types
+
+import           UnliftIO.Async
+import           UnliftIO.STM
+
+-- | Implementation of the 'CommitSmartBuffer' strategy: We use an
+-- async timer for committing cursors at specified intervals, but if
+-- the number of uncommitted events reaches some threshold before the
```
erthalion: Since it can be considered as documentation, I would rephrase this part. Instead of "some threshold", "the threshold, specified by maxUncommittedEvents, divided by 2".

mtesseract: done

src/Network/Nakadi/Internal/Committer.hs
```diff
@@ -0,0 +1,74 @@
...
+-- all cursors synchronously, one after the other.
+subscriptionSink
+  :: (MonadIO m, MonadNakadi b m)
+  => SubscriptionEventStream
+  -> ConduitM (SubscriptionEventStreamBatch a) void m ()
+subscriptionSink eventStream = do
+  config <- lift nakadiAsk
+  awaitForever $ \batch -> lift $ do
+    let cursor = batch ^. L.cursor
+    catchAny (commitOneCursor eventStream cursor) $ \exn ->
```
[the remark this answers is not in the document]
mtesseract: Good catch! I have removed the exception catching here. Also, this removes log duplication.
Thanks!

Regarding the propagation of errors during committing:

In the past my team has experienced that Nakadi works better (i.e., with less waiting time between reconnects) when the consumer does not forcefully disconnect on committing failures, but instead continues committing as far as it can an then let the Nakadi server do the disconnect when it thinks that it is required. This way Nakadi is in control.

It's up to debate, I would say, whether this is still the best approach. But, as far as I know, the current nakadi-client version does it like this, thus I would prefer to do any modifications regarding this in a seperate issue/PR.

Does this answer all your questions?

src/Network/Nakadi/Internal/Worker.hs
```diff
@@ -0,0 +1,199 @@
...
+          -- lookup failures.
+          NonEmpty.head workers
+        Just idx ->
+          -- Lookup successful. Truncate the resulting index using modulo in order to
+          -- obtain a worker index and return the corresponding worker reference.
+          workers NonEmpty.!! (idx `mod` nWorkers)
+
+-- | Block as long no worker has finished. Workers are supposed to run forever, unless
+-- cancelled. Therefore, using 'waitAnyCancel' essentially means that if some worker
+-- fails due to an uncaught exception, then all other workers are cancelled as well.
```
erthalion: But why? I mean, is it something useful from your own experience?

erthalion:
Other than my last question, I think I reached my limits in reviewing of this PR, so :+1:

src/Network/Nakadi/Internal/Worker.hs
erthalion: Well, you've got the point. But I more thought about the situation when an error was caused by the data that was sent to a worker - in this case I'm not sure if it makes sense to stop everything, but at the same time all other strategies sound too much over complicated. So probably the current implementation is indeed a better solution.

erthalion:
@mtesseract yep, sure!

src/Network/Nakadi/Internal/Worker.hs
mtesseract: We should indeed make sure that the workers themselves are resilient against certain form of processing errors!

PR #119, opened 2018-07-13
Implement MonadNakadi for IO using global configuration

### LINKED ISSUE #117: Implement MonadNakadi for IO monad
This enables new use-cases: quick and easy testing / usage of nakadi-client API e.g. in GHCi.
Could be interesting for writing scripts for Nakadi.
Also, would enable quick boot strapping of nakadi-client.

Using some form of reader monad in a custom monad stack or `NakadiT` is still the recommended way of integrating the library, though.

mtesseract:
Implement MonadNakadi for IO.
Include functions for producing configuration from environment.

This simplifies quick testing in the REPL and enables simple scripting, without monad transformers.

closes #117 

@etorreborre, would you be available for a review?

### DIFF
--- a/src/Network/Nakadi/Internal/Types.hs
+++ b/src/Network/Nakadi/Internal/Types.hs
@@ -12,16 +12,14 @@ Exports all types for internal usage.
 
 {-# LANGUAGE ConstraintKinds            #-}
 {-# LANGUAGE DefaultSignatures          #-}
-{-# LANGUAGE DeriveFunctor              #-}
 {-# LANGUAGE FlexibleContexts           #-}
 {-# LANGUAGE FlexibleInstances          #-}
 {-# LANGUAGE FunctionalDependencies     #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE MultiParamTypeClasses      #-}
 {-# LANGUAGE RankNTypes                 #-}
-{-# LANGUAGE RecordWildCards            #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE TypeFamilies               #-}
+{-# LANGUAGE LambdaCase                 #-}
 {-# LANGUAGE UndecidableInstances       #-}
 
 module Network.Nakadi.Internal.Types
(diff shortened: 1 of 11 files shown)


src/Network/Nakadi/Internal/Types.hs
```diff
@@ -195,6 +196,14 @@ instance MonadBaseControl b' m => MonadBaseControl b' (NakadiT b m) where
 -- | 'MonadNakadiBase'
 instance {-# OVERLAPPABLE #-} MonadNakadiBase b m => MonadNakadiBase b (NakadiT b m)
 
+-- ** Implementation of 'MonadNakadi' for IO. Useful for testing purposes or scripting.
+
+instance MonadNakadi IO IO where
+  nakadiAsk =
+    liftIO (readTVarIO globalConfiguration) >>= \case
```
etorreborre: Is this still going to work if there are `local` actions where the config might be locally modified? If you `nakadiAsk` the configuration wouldn't you get the global config all the time?

etorreborre: Overall I am wondering what we can not accomplish with `NakadiT` and maybe a `runNakadiTIO` function creating a default config and running a `NakadiT IO IO` action?

mtesseract: Regarding `local`:

There is no `local` if you are in `IO`. If you have such requirements (in most applications you most likely have them!) then you should of course go with the recommended setup: use a monad transformer stack including some kind of reader monad, which then provides `local` functionality. This could be either `NakadiT` or some custom monad monad that implemens `MonadNakadi`.

The signature of `local` is:

```
local :: MonadReader r m => (r -> r) -> m a -> m a
```

As you can see, if `m == IO`, this would not type check.

mtesseract: Regarding `NakadiT`:

What this PR allows is to use Nakadi on the REPL directly, as in e.g.:

```
ghci> subscriptions <- subscriptionList
ghci> mySubscriptions = filter isItMine subscriptions
ghci> forM_ mySubscriptions (subscriptionDelete . view L.id)
```

The whole purpose is to not sprinkle `runNakadiT` (or `runNakadiTIO`) all over the code.

I am imaging scripting against Nakadi to a possible use-case of this. Having a `MonadNakadi` instance for `IO` would simply be a bonus for ergonomics in case you don't need a full-featured reader monad.

etorreborre: I get it. I was thinking that `MonadNakadi` by itself would allow programs to use `MonadReader` and then `local`. I think this is indeed useful for scripting, even if I am super reluctant of having some global mutable state :-). But since this is in a separate module which I might even call `Network.Nakadi.Unsafe.IO` that should be fine :-).

mtesseract:
We can use your suggestion for the module name.
I would not consider it to be *that* unsafe, though. :-)
Cool, I will update the PR (just changing the module name) and go ahead with merging if that is fine for you!

mtesseract:
Ok, I have added a few more changes:

* Renamed the module
* Exposed the public one in package.yaml
* Added two tests for this IO interface

Neither changes to the actual instance definition, nor to the definition of the two functions for setting the global config.

Will merge!
Thanks!

PR #128, opened 2018-09-23
Fixes haddock building

taojang:
* Fix indentation, otherwise downstream code uses the library won't be able to create local hoogle db
* Add additional step to run `stack haddock` to CI script

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,5 +47,4 @@ after_script:
 - (cd $HOME/nakadi; TERM=dumb ./gradlew stopNakadi)
 
 script:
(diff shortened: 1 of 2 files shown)


taojang:
I'm sorry I think @etorreborre has already fixed the issue of haddock failing. We are experiencing this because of using a outdated fork. But I would like to add haddock into part of CI test to make sure this won't happen. Hope it still runs within the maximum allowed running time on travis :pray:

mtesseract:
Hi! Cool, thank you very much for this.
Shows that the haddock documentation needs quite some improvement — so many unresolved links, for example.

PR #132, opened 2019-01-05
Update docker-compose on Travis

mtesseract:


mtesseract:
This PR just updates docker-compose. But the test suite is broken due a different bug, which will be resolved in a separate PR (https://github.com/mtesseract/nakadi-client/issues/133).

Thus, merging.

PR #134, opened 2019-01-07
Rework implementation of Problem type.

mtesseract:
The old implementation was in violation of RFC7807.
Use new dependency for type-safe encoding of URIs: modern-uri.
Added tests.

**This contains breaking API changes related to the `Problem` type.**

Therefore, version bump to 0.7.0.0.

Could anyone of you do a review please?
@taojang @amrhassan @rbobin @erthalion
(Or direct me to other users of the package? I am currently on parental leave. Thanks!)


mtesseract:
Background: I was adding support for the configurable commit timeout, but it turned out that the current master branch fails due to several reasons: (1) outdated docker-compose on Travis -- fixed and (2) deserialisation failures when running the integration test suite against recent Nakadi server versions.

When looking into the problem I have realised that the implementation of the Problem type (RFC7807) was flawed in Nakadi-client. In this PR I hope to fix it.

mtesseract:
Since there was no feedback on this one, I will proceed with merging.

